PR #1, opened 2018-07-14
Added support for openssl 1.1. Removed gtest src dependencies

galuha:
- Added support for openssl 1.1.
- Removed gtest src dependencies, that caused unnecessary make editing
- Removed std=c++14 dependencie in order to support older compilers

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -732,7 +732,7 @@ namespace jwt {
 	}
 
 	struct default_clock {
-		std::chrono::_V2::system_clock::time_point now() const {
+		std::chrono::system_clock::time_point now() const {
 			return std::chrono::system_clock::now();
 		}
 	};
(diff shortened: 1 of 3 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -688,12 +731,12 @@ namespace jwt {
...
 	}
 
-	auto verify() {
-		struct default_clock {
-			auto now() const {
-				return std::chrono::system_clock::now();
-			}
-		};
+	struct default_clock {
+		std::chrono::_V2::system_clock::time_point now() const {
```
Thalhammer: Shouldn't this be std::chrono::system_clock::time_point ?

Otherwise thank you for your contribution, will merge as soon as possible.

galuha: Fixed. And thank you for this lightweight library

Thalhammer:
Did some improvements in e162b97f5b03bf59ae51c9fc54f5952124d93b60 to support older compilers and keep the tests clean, but otherwise good work.

PR #3, opened 2018-08-02
Fix warnings in JWT

ThalmicRevane:
Fix some warnings:
* Extra braces for braced-initialization using initializer_list
* Unused parameters.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -70,10 +70,10 @@ namespace jwt {
 	namespace algorithm {
 
 		struct none {
-			std::string sign(const std::string& data) const {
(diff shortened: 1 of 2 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -70,10 +70,10 @@ namespace jwt {
 	namespace algorithm {
 
 		struct none {
-			std::string sign(const std::string& data) const {
+			std::string sign(const std::string&) const {
```
ThalmicRevane: Would you prefer I just comment out the name instead of removing it?

Thalhammer: I think it is fine to remove it.

PR #4, opened 2018-08-02
Mark global functions inline

ThalmicRevane:
Without this change including the jwt header in multiple translation units will
end up with ODR violations.


Thalhammer:
This was indeed an error

PR #22, opened 2019-04-29
Vcpkg support

webfolderio:


Thalhammer:
Looks good, but that will break standalone builds, as the picojson.h is no more in included correctly.
Mind moving it into the picojson subfolder so building outside of vcpkg still works ?

webfolderio:
> but that will break standalone builds

This pull request do not touch existing building file.

> Mind moving it into the picojson subfolder so building outside of vcpkg still works ?

No needed, existing build still works.

Here is the steps to build **jwt-cpp** with **vcpg**.

```
git clone https://github.com/Thalhammer/jwt-cpp
cd jwt-cpp
mkdir c:\tools\vcpkg\ports\jwt-cpp
copy vcpkg\*.* c:\tools\vcpkg\ports\jwt-cpp
vcpkg install picojson:x64-windows
vcpkg install jwt-cpp:x64-windows

```

Thalhammer:
Nevermind, I'm stupid.
I thought the patch was a change in the actuall jwt.h

PR #23, opened 2019-05-02
Fix checking for fallthrough attribute support for older compilers

martin-flaska:
GCC 4.9 (tested on 4.9.3)  doesn't have __has_cpp_attribute.

Compilation fails with:

```
include/jwt-cpp/jwt.h:1139:24: error: missing binary operator before token "("
 #if __has_cpp_attribute(fallthrough)
                        ^
```


Thalhammer:
Looks good, thanks for your contribution.

PR #29, opened 2019-05-05
typecast to avoid errors

### LINKED ISSUE #27: Compilation errors - related to openssl version?
Good day, I'm still investigating this, but curious if you are aware of typecasting errors of this library and issues related to openssl version, possibly? 

I am able to build your library and use it just fine, but I'm trying to integrate this into a 3rd party system and the build is failing on travis there. I am currently not sure if their VM is using different versions of libraries, but its on ubuntu (as am I).

I'll be digging in more, but if the error looks familiar, would love a hint. Maybe [this fix](https://github.com/eidheim/Simple-Web-Server/issues/150) is related?

```
../third_party/jwt-cpp/include/jwt-cpp/jwt.h: In function ‘std::string jwt::helper::extract_pubkey_from_cert(const string&, const string&)’:
../third_party/jwt-cpp/include/jwt-cpp/jwt.h:77:87: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
    std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf(certstr.data(), certstr.size()), BIO_free_all);
                                                                                       ^
In file included from /usr/include/openssl/evp.h:75:0,
                 from ../third_party/jwt-cpp/include/jwt-cpp/jwt.h:9,
                 from src/zm_crypt.h:27,
                 from /src/zm_user.cpp:32:
/usr/include/openssl/bio.h:668:6: error:   initializing argument 1 of ‘BIO* BIO_new_mem_buf(void*, int)’ [-fpermissive]
 BIO *BIO_new_mem_buf(void *buf, int len);
      ^
```

pliablepixels:
fixes #27

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -74,13 +74,19 @@ namespace jwt {
 	namespace helper {
 		inline
 		std::string extract_pubkey_from_cert(const std::string& certstr, const std::string& pw = "") {
-			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf((char *)certstr.data(), certstr.size()), BIO_free_all);
+			// TODO: Cannot find the exact version this change happended
+#if OPENSSL_VERSION_NUMBER <= 0x1000114fL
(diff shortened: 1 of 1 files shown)


Thalhammer:
I added a ifdef and changed the raw cast to a const_cast.
As well as fixed a false positive detected by codacy.

pliablepixels:
thank you!

PR #54, opened 2020-01-21
Update project to use cmake ( and vcpkg )

prince-chrismc:
I wanted to include this in my project tree, which is cmake based. I updated the code base to use this common tool, share the work since I am sure others will appreciate it as well.

Since the is meant to be used with vcpkg, I used it to install the dependencies and configure the project.

Feel free to provide feedback =)


CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+set(JWT_INCLUDE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_include_directories(
+  jwt-cpp INTERFACE $<BUILD_INTERFACE:${JWT_INCLUDE_PATH}>
+                    $<INSTALL_INTERFACE:include>)
+
+target_link_libraries(jwt-cpp INTERFACE OpenSSL::SSL OpenSSL::Crypto)
+
+set(JWT_HEADERS ${JWT_INCLUDE_PATH}/jwt-cpp/base.h
+                ${JWT_INCLUDE_PATH}/jwt-cpp/jwt.h)
+set(PICO_HEADER ${JWT_INCLUDE_PATH}/picojson/picojson.h)
```
prince-chrismc: picojson has poor cmake support so it's most convenient to keep a copy

vcpkg/portfile.cmake
```diff
@@ -1,23 +1,23 @@
...
 include(vcpkg_common_functions)
 
-set(SOURCE_PATH ${CURRENT_BUILDTREES_DIR}/src/jwt-cpp)
-
-vcpkg_from_github(OUT_SOURCE_PATH SOURCE_PATH
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
     REPO Thalhammer/jwt-cpp
     REF f0e37a79f605312686065405dd720fc197cc3df0
     SHA512 ae83c205dbb340dedc58d0d3f0e2453c4edcf5ce43b401f49d02692dc8a2a4b7260f1ced05ddfa7c1d5d6f92446e232629ddbdf67a58a119b50c5c8163591598
```
prince-chrismc: This needs to be updated and tested before merging!

Thalhammer:
> Since the is meant to be used with vcpkg, I used it to install the dependencies and configure the project.

Sorry for the late reply, I was quite busy with other stuff.
The Project is not really "meant" to be used with vcpkg, it is just a convenience. I dont use it at all. Since the library is header only and mostly selfcontained you should just need to checkout it out and append the include directory to your path. There is nothing that needs to be built, all .cpp files are just unit tests.

prince-chrismc:
The main goal was to use cmake so it can more easily be included with versioning semantics and setting. (And for me to practice)

If you are not interested, you may close the PR =)

Thalhammer:
@prince-chrismc Is there a reason why you decided to build openssl instead of using the ubuntu provided one or is this just personal preference ?

Because it increases the build time by a lot and since we do not ship the generated binaries it does not matter if they have a specific version.

prince-chrismc:
The default is 1.0.2 for openssl, I don't remember 100% but there was problems when using lower then 1.1... The make would also need to be updated with the correct version

Thalhammer:
Thanks for your work, proper cmake support was long overdue.

prince-chrismc:
Always a pleasure. Your library has been a delight! Thank you for your work =)

Once https://github.com/conan-io/conan-center-index/pull/1031 is merged, I'll be able to use it much more =)

PR #59, opened 2020-03-24
Adding stream operators to jwt::claim

prince-chrismc:
I also removed some error prone code duplication with the `explicit` keyword for claims


Thalhammer:
Looks good to me.

PR #71, opened 2020-06-06
A generic basic_claim to allow support for other libraries

prince-chrismc:
> _In full disclosure, generic template meta-programming containers has been a learning curve_ 

Take a look, certainly can use feedback on this. Maybe it's not something that fits this project :man_shrugging: 

Currently I am using 3 JSON libraries, and have 3 base64 implementations so having the ability to switch out the the underlying library would be a huge asset. Others have asked for this so I went ahead with my best attempt.

Open question, should there be official nlohmann/json support?

I need to address the CMake I would like to add a few options to exclude `picojson.h` and `base.h` since it will be possible to no use those.

Another point to look at, perhaps user feedback, the number of template parameters. Its very possible to have a single one `traits` which has all the type definitions required.

I this gets merged it would be nice to have a pre-release and post it online for feedback :crossed_fingers: to get this over the hump.

One goal was to not break the API, there should be zero changes that affect any consumer. there's only one change in the existing test code


Thalhammer:
I'll take a look at it tomorrow, but I certainly like the idea of having base64/json libraries exchangeable.
The base64 is not really visible on any api so thats probably less of a priority, but the json data hits user code, so it would be a huge benefit to support a library they already know.
When I initially wrote this library the goal was to support a personal project and since I use picojson everywhere and nothing else It was never a design consideration.

prince-chrismc:
The last open question is do we official support nlohmann/json?

Short of that I think I've checked all the boxes. When you get the chance, love for a code review :smiley:

tests/NlohmannTest.cpp
```diff
@@ -0,0 +1,151 @@
...
+	nlohmann::json::number_integer_t, \
+	nlohmann::json::boolean_t, nlohmann_traits
+
+TEST(NholmannTest, BasicClaims) {
+	using nholmann_claim =
+			jwt::basic_claim<JWT_NHOLMANN_CLAIM_TPL>;
+
+	const auto string = nholmann_claim(std::string("string"));
+	const auto array = nholmann_claim(std::set<std::string>{"string", "string"});
+	const auto integer = nholmann_claim(159816816);
```
prince-chrismc, line 88: nlohmann not nholmann

Thalhammer:
I am sorry it took me this long to take a proper look at the amazing work you did here.
I do like what I see, however I have a couple of minor things I would do differently.
* You seem to import a couple of C++14/C++17 features.
   There are feature test macros which are defined by the compiler if a given feature is provided by the stdlib ([See here](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B17).
   While not a big deal, it might be a good idea to use the "official" one if supported.
* I honestly dont like macros so I might be a bit biased, but I don't see a reason not to move the types into the json trait and passing only the traits around, instead of using a macro to add them. Is there a reason why you did it this way ?

None of the above are a big deal/show stopper, though.

prince-chrismc:
That's absolutely all right, one thing i learnt is every one is too busy!

>  You seem to import a couple of C++14/C++17 features.
>  There are feature test macros which are defined by the compiler if a given feature is provided by the stdlib ([See here](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B17)).
>   While not a big deal, it might be a good idea to use the "official" one if supported.

Totally in agreement.

The only feature I ~implemented~copied myself was [std::is_detected](https://en.cppreference.com/w/cpp/experimental/is_detected) which is not even apart of the standard, [__cpp_lib_experimental_detect](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#v2), I'll add it to my agenda for the next rainy weekend :smile: 

If there are any other's let me know.

> I honestly dont like macros so I might be a bit biased, but I don't see a reason not to move the types into the json trait and passing only the traits around, instead of using a macro to add them. Is there a reason why you did it this way ?

I was following the nlohmann/json implementation. I had been considering it myself. It was in my [opening comment](https://github.com/Thalhammer/jwt-cpp/pull/71#issue-428900468)!

Thalhammer:
> If there are any other's let me know.

make_void was the one that jumped my eye (which is c++17), but I honestly dont know the standards after C++14 well enough to judge what else might get replaced.


> It was in my [opening comment](https://github.com/Thalhammer/jwt-cpp/pull/71#issue-428900468)!

Sorry missed that :)

prince-chrismc:
I think all the possible features are int eh same block so it should not be too hard to check them. __cpp_lib_void_t might cover make_void

Issue #41, opened 2019-09-17
Api referece - documentation


nicraMarcin:
Where can I find some api referece and some examples? for example: how to use with rsa public and private key

Thalhammer:
Currently the only documentation are the example in README.md and the various testcases in *Test.cpp

nicraMarcin:
:(

Thalhammer:
I almost forgot about it, but most of the methods are commented with doxygen comments and there is a doxygen config in the root. It's not really a complete documentation, but it might help.

paulotovo:
Hello Thalhammer,

I have seen on several sites about the header "typ" parameter that must be configured as JWT.
For example: "typ" : "JWT"
But on RFC says:
"The "typ" (type) Header Parameter defined by [JWS] and [JWE] is used
   by JWT applications" and " If present, it is RECOMMENDED that
   its value be "JWT" to indicate that this object is a JWT."

In your examples you use "typ" : "JWS".

What would be the difference? What would be the correct use?

Thanks.

Best Regards

Paulo

Thalhammer:
@paulotovo According to the rfc that part is optional and should not be used by the jwt library.
It defines the media type and could be used to disabiguate if you expect not only jwt tokens to hit your app. JWT stands for Json Web Tokens and JWS for Json Web Signature which is a different, but very similar rfc. In 99% of cases it does not matter and the reason my test token uses JWS is simply a coincidence.

minjax:
Continuing with the theme of API documentation/examples, the example given shows get_payload_claims() in this context:
```
      std::string token = "eyJhbGciOiJIUzI1...OkEE";
      auto decoded = jwt::decode(token);

      for( auto& e : decoded.get_payload_claims() )
          std::cout << e.first << " = " << e.second.to_json() << std::endl;
```
but, I don't find any definition of get_payload_claims() in class jwt::decoded_jwt<jwt::picojson_traits> or otherwise...

What am I missing here?

Thanks,

prince-chrismc:
The current master is a getting ready for an release candidate, and rightfully so because it seem like `get_payload_claims` got removed during #71 by accident :fearful: 

I would **strongly** recommend using the v0.4.0 release while the last few details get ironed out. ( I am using that in production code ) The method in question is available there.

https://github.com/Thalhammer/jwt-cpp/blob/34bb0644ea613cfcbc09c148db9de8aa6c5612b5/include/jwt-cpp/jwt.h#L1027

[...]

PR #75, opened 2020-07-01
Compile examples

prince-chrismc:
There was an API break introduced by #71 which makes the example in the README not compile \cc minjax

This will add the examples requested in #41 as well as test:
- the examples compile
- the cmake `find_package` is correct

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,7 @@ script:
   - cmake .
   - make
   - ./tests/jwt-cpp-test
+  - make install
+  - cd example
+  - cmake .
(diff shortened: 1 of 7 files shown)


prince-chrismc:
While testing the example 
```cpp
auto token = jwt::create()
	.set_issuer("auth0")
	.set_issued_at(std::chrono::system_clock::now())
	.set_expires_at(std::chrono::system_clock::now() + std::chrono::seconds{3600})
	.sign(jwt::algorithm::hs256{"secret"});
```
is also broken

PR #86, opened 2020-07-16
Initial support for std::error_code

Thalhammer:
Initial support for error_code overloads.
@prince-chrismc Wanna take a quick look to check if I maybe missed something ?

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -533,25 +533,9 @@ namespace jwt {
 			/**
 			 * \brief Return an empty string
 			 */ 
-			std::string sign(const std::string& /*unused*/) const {
-				return {};
-			}
-			/**
-			 * \brief Return an empty string
-			 */ 
 			std::string sign(const std::string& /*unused*/, std::error_code& ec) const {
 				ec.clear();
 				return {};
-			}
-			/**
-			 * \brief Check if the given signature is empty.
-			 * 
-			 * JWT's with "none" algorithm should not contain a signature.
-			 * \throw signature_verification_exception
-			 */ 
-			void verify(const std::string& /*unused*/, const std::string& signature) const {
-				if (!signature.empty())
-					throw signature_verification_exception();
 			}
 			/**
 			 * \brief Check if the given signature is empty.
@@ -587,18 +571,6 @@ namespace jwt {
 			/**
 			 * Sign jwt data
 			 * \param data The data to sign
-			 * \return HMAC signature for the given data
-			 * \throw signature_generation_exception
-			 */
-			std::string sign(const std::string& data) const {
-				std::error_code ec;
-				auto res = sign(data, ec);
-				error::throw_if_error(ec);
-				return res;
-			}
-			/**
-			 * Sign jwt data
-			 * \param data The data to sign
 			 * \param ec error_code filled with details on error
 			 * \return HMAC signature for the given data
 			 */
@@ -618,36 +590,21 @@ namespace jwt {
 			 * Check if signature is valid
 			 * \param data The data to check signature against
 			 * \param signature Signature provided by the jwt
-			 * \throw signature_verification_exception If the provided signature does not match
-			 */
-			void verify(const std::string& data, const std::string& signature) const {
-				std::error_code ec;
-				verify(data, signature, ec);
-				error::throw_if_error(ec);
-			}
-			/**
-			 * Check if signature is valid
-			 * \param data The data to check signature against
-			 * \param signature Signature provided by the jwt
 			 * \param ec Filled with details about failure.
 			 */
 			void verify(const std::string& data, const std::string& signature, std::error_code& ec) const {
 				ec.clear();
-				try {
-					auto res = sign(data);
-					bool matched = true;
-					for (size_t i = 0; i < std::min<size_t>(res.size(), signature.size()); i++)
-						if (res[i] != signature[i])
-							matched = false;
-					if (res.size() != signature.size())
+				auto res = sign(data, ec);
+				if(ec) return;
+				bool matched = true;
+				for (size_t i = 0; i < std::min<size_t>(res.size(), signature.size()); i++)
+					if (res[i] != signature[i])
 						matched = false;
-					if (!matched) {
-						ec = error::signature_verification_error::invalid_signature;
-						return;
-					}
-				}
-				catch (const signature_generation_exception&) {
+				if (res.size() != signature.size())
+					matched = false;
+				if (!matched) {
 					ec = error::signature_verification_error::invalid_signature;
+					return;
 				}
 			}
 			/**
@@ -691,18 +648,6 @@ namespace jwt {
 			/**
 			 * Sign jwt data
 			 * \param data The data to sign
-			 * \return RSA signature for the given data
-			 * \throw signature_generation_exception
-			 */
-			std::string sign(const std::string& data) const {
-				std::error_code ec;
-				auto res = sign(data, ec);
-				error::throw_if_error(ec);
-				return res;
-			}
-			/**
-			 * Sign jwt data
-			 * \param data The data to sign
 			 * \param ec error_code filled with details on error
 			 * \return RSA signature for the given data
 			 */
@@ -737,17 +682,6 @@ namespace jwt {
 
 				res.resize(len);
 				return res;
-			}
-			/**
-			 * Check if signature is valid
-			 * \param data The data to check signature against
-			 * \param signature Signature provided by the jwt
-			 * \throw signature_verification_exception If the provided signature does not match
-			 */
-			void verify(const std::string& data, const std::string& signature) const {
-				std::error_code ec;
-				verify(data, signature, ec);
-				error::throw_if_error(ec);
 			}
 			/**
 			 * Check if signature is valid
@@ -855,18 +789,6 @@ namespace jwt {
 			/**
 			 * Sign jwt data
 			 * \param data The data to sign
-			 * \return ECDSA signature for the given data
-			 * \throw signature_generation_exception
-			 */
-			std::string sign(const std::string& data) const {
-				std::error_code ec;
-				auto res = sign(data, ec);
-				error::throw_if_error(ec);
-				return res;
-			}
-			/**
-			 * Sign jwt data
-			 * \param data The data to sign
 			 * \param ec error_code filled with details on error
 			 * \return ECDSA signature for the given data
 			 */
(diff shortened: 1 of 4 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -107,58 +307,165 @@ namespace jwt {
...
+			ec.clear();
 #if OPENSSL_VERSION_NUMBER <= 0x10100003L
 			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf(const_cast<char*>(certstr.data()), certstr.size()), BIO_free_all);
 #else
 			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf(certstr.data(), static_cast<int>(certstr.size())), BIO_free_all);
 #endif
 			std::unique_ptr<BIO, decltype(&BIO_free_all)> keybio(BIO_new(BIO_s_mem()), BIO_free_all);
+			if(!certbio || !keybio) {
+				ec = error::rsa_error::create_mem_bio_failed;
+				return "";
```
prince-chrismc, line 328:
```suggestion
				return {};
```

better optimization https://godbolt.org/z/93Phdq

include/jwt-cpp/jwt.h
```diff
@@ -107,58 +307,165 @@ namespace jwt {
...
 			std::string res(ptr, len);
 			return res;
 		}
 
+		/**
+		 * \brief Extract the public key of a pem certificate
+		 * 
+		 * \param certstr	String containing the certificate encoded as pem
+		 * \param pw		Password used to decrypt certificate (leave empty if not encrypted)
+		 * \throws	rsa_exception if an error occurred
```
prince-chrismc, line 360:
```suggestion
		 * \throws		rsa_exception if an error occurred
```

include/jwt-cpp/jwt.h
```diff
@@ -107,58 +307,165 @@ namespace jwt {
...
 		}
 
+		/**
+		 * \brief Load a public key from a string.
+		 * 
+		 * The string should contain a pem encoded certificate or public key
+		 * 
+		 * \param certstr	String containing the certificate or key encoded as pem
+		 * \param pw		Password used to decrypt certificate or key (leave empty if not encrypted)
+		 * \throws	rsa_exception if an error occurred
```
prince-chrismc, line 418:
```suggestion
		 * \throws		rsa_exception if an error occurred
```

include/jwt-cpp/jwt.h
```diff
@@ -209,6 +516,13 @@ namespace jwt {
 			std::string sign(const std::string& /*unused*/) const {
 				return "";
```
prince-chrismc, line 517:
```suggestion
				return {};
```

include/jwt-cpp/jwt.h
```diff
@@ -209,6 +516,13 @@ namespace jwt {
...
 			}
+			/**
+			 * \brief Return an empty string
+			 */ 
+			std::string sign(const std::string& /*unused*/, std::error_code& ec) const {
+				ec.clear();
+				return "";
```
prince-chrismc, line 524:
```suggestion
				return {};
```

include/jwt-cpp/jwt.h
```diff
@@ -219,6 +533,19 @@ namespace jwt {
 				if (!signature.empty())
 					throw signature_verification_exception();
 			}
+			/**
+			 * \brief Check if the given signature is empty.
+			 * 
+			 * JWT's with "none" algorithm should not contain a signature.
+			 * \param signature Signature data to verify
```
prince-chrismc, line 540:
```suggestion
			 * \param signature		Signature data to verify
```

include/jwt-cpp/jwt.h
```diff
@@ -244,11 +571,26 @@ namespace jwt {
...
+			std::string sign(const std::string& data, std::error_code& ec) const {
+				ec.clear();
 				std::string res;
 				res.resize(static_cast<size_t>(EVP_MAX_MD_SIZE));
 				auto len = static_cast<unsigned int>(res.size());
-				if (HMAC(md(), secret.data(), static_cast<int>(secret.size()), reinterpret_cast<const unsigned char*>(data.data()), static_cast<int>(data.size()), (unsigned char*)res.data(), &len) == nullptr)  // NOLINT(google-readability-casting) requires `const_cast`
-					throw signature_generation_exception();
+				if (HMAC(md(), secret.data(), static_cast<int>(secret.size()), reinterpret_cast<const unsigned char*>(data.data()), static_cast<int>(data.size()), (unsigned char*)res.data(), &len) == nullptr) { // NOLINT(google-readability-casting) requires `const_cast` 
+					ec = error::signature_generation_error::hmac_failed;
+					return "";
```
prince-chrismc, line 592:
```suggestion
					return {};
```

include/jwt-cpp/jwt.h
```diff
@@ -107,58 +307,165 @@ namespace jwt {
...
+		std::string extract_pubkey_from_cert(const std::string& certstr, const std::string& pw, std::error_code& ec) {
+			ec.clear();
 #if OPENSSL_VERSION_NUMBER <= 0x10100003L
 			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf(const_cast<char*>(certstr.data()), certstr.size()), BIO_free_all);
 #else
 			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new_mem_buf(certstr.data(), static_cast<int>(certstr.size())), BIO_free_all);
 #endif
 			std::unique_ptr<BIO, decltype(&BIO_free_all)> keybio(BIO_new(BIO_s_mem()), BIO_free_all);
+			if(!certbio || !keybio) {
+				ec = error::rsa_error::create_mem_bio_failed;
```
prince-chrismc, line 327: could be more generic than `rsa_error`

All the headers return an `rsa_error`, do any of the other algos use the helpers? I don't see any in the diff

include/jwt-cpp/jwt.h
```diff
@@ -99,6 +99,206 @@ namespace jwt {
...
+					case signature_generation_error::signinit_failed: return "failed to create signature: SignInit failed";
+					case signature_generation_error::signupdate_failed: return "failed to create signature: SignUpdate failed";
+					case signature_generation_error::signfinal_failed: return "failed to create signature: SignFinal failed";
+					case signature_generation_error::ecdsa_do_sign_failed: return "failed to generate ecdsa signature";
+					case signature_generation_error::digestinit_failed: return "failed to create signature: DigestInit failed";
+					case signature_generation_error::digestupdate_failed: return "failed to create signature: DigestUpdate failed";
+					case signature_generation_error::digestfinal_failed: return "failed to create signature: DigestFinal failed";
+					case signature_generation_error::rsa_padding_failed: return "failed to create signature: RSA_padding_add_PKCS1_PSS_mgf1 failed";
+					case signature_generation_error::rsa_private_encrypt_failed: return "failed to create signature: RSA_private_encrypt failed";
+					default: return "unknown error code";
```
prince-chrismc, line 229:
```suggestion
					default: return "unknown signature generation error";
```

include/jwt-cpp/jwt.h
prince-chrismc: This could be applied throughout, maybe a separate PR?

include/jwt-cpp/jwt.h
```diff
@@ -99,6 +99,206 @@ namespace jwt {
...
+				std::string message(int ev) const override {
+					switch(static_cast<token_verification_error>(ev)) {
+					case token_verification_error::ok: return "no error";
+					case token_verification_error::wrong_algorithm: return "wrong algorithm";
+					case token_verification_error::missing_claim: return "decoded_jwt is missing required claim";
+					case token_verification_error::claim_type_missmatch: return "claim type does not match expected type";
+					case token_verification_error::claim_value_missmatch: return "claim value does not match expected value";
+					case token_verification_error::token_expired: return "token expired";
+					case token_verification_error::audience_missmatch: return "token doesn't contain the required audience";
+					default: return "unknown error code";
```
prince-chrismc, line 264:
```suggestion
					default: return "unknown token verification error";
```

include/jwt-cpp/jwt.h
```diff
@@ -99,6 +99,206 @@ namespace jwt {
...
+		inline std::error_category& token_verification_error_category() {
+			class token_verification_error_cat : public std::error_category
+			{
+			public:
+				const char* name() const noexcept override { return "token_verification_error"; };
+				std::string message(int ev) const override {
+					switch(static_cast<token_verification_error>(ev)) {
+					case token_verification_error::ok: return "no error";
+					case token_verification_error::wrong_algorithm: return "wrong algorithm";
+					case token_verification_error::missing_claim: return "decoded_jwt is missing required claim";
```
prince-chrismc, line 259:
```suggestion
					case token_verification_error::missing_claim: return "decoded JWT is missing required claim(s)";
```

I noticed the other error codes return "failed to ...." this one does not, in case you want them too.

Thalhammer:
> Thank you for the invite firstly! I want to help make this project successful 

Well since you already touched/wrote like half of the codebase and I never had any real complaints I think thats just fair.

> I noticed the exceptions and now the error_codes for the "algorithms" are outside of that namespace, would you think they could be moved to that namespace?

I though about that, but if you have catch blocks moving it means users have to change their code. But thinking about it we have breaking changes for the next release anyway.

prince-chrismc:
> I though about that, but if you have catch blocks moving it means users have to change their code.

:100: it does not add to the UX either, 

>  But thinking about it we have breaking changes for the next release anyway.

I don't actually see any API breaks, it might be worth reviewing to make sure. Very little test code changed.

> I think its a good thing to keep all the error stuff together in its own namespace, rather than putting it in the algorithm namespace.

I am certainly in favor of that!

Thalhammer:
Maybe we could move all exception types to the error namespace and just typedef them in the jwt namespace ?

prince-chrismc:
> Maybe we could move all exception types to the error namespace and just typedef them in the jwt namespace ?

```cpp
namespace error
{
    //rsa_expection...
    //ecdsa_expection...
}

using rsa_exception;
using ecdsa_expection;
```

:thinking: 

That would be a great alternative!

prince-chrismc:
Looking at the coverage the biggest missed count goes to failed openssl calls. which is not exactly easy to test.

Perhaps a few of them could be marks with [these](https://stackoverflow.com/a/23264375/8480874) so they are not counted

Thalhammer:
> Looking at the coverage the biggest missed count goes to failed openssl calls. which is not exactly easy to test.

I already have a unit test in the working to test those. However it's a bit hacky, so I'd like to clean it up a bit more and do it in a different PR. Basically what I do is I override all OpenSSL functions we use and have a bitmask which decides when it should fail. If it should fail I return one of the error codes as documented by Openssl and if not I call the original function. However this method only works on linux, which is fine I guess ?

prince-chrismc:
> However this method only works on linux, which is fine I guess ?

Absolutely, if we add windows tests, that can always be `#ifndef _WIN32` to remove them

Thalhammer:
@prince-chrismc Think I can merge this ?

PR #91, opened 2020-07-22
Fixed a build error

Sp3EdeR:
`jwt.h` previously depended on the nlohmann JSON library's inclusion of `iso646.h`, which defines the `not` operator.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -1384,7 +1384,7 @@ namespace jwt {
 		struct supports_as_integer {
 			static constexpr auto value =
 				std::is_signed<integer_type>::value &&
-				not std::is_floating_point<integer_type>::value &&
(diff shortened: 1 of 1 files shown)


Sp3EdeR:
The MSVC compilers do not recognize the alternative operator representations, and iso646.h is not necessarily included.

prince-chrismc:
Windows CI?

Sp3EdeR:
Cross-platform CI actually. But I usually start with the Windows stuff on open source libraries, since that is what they're least tested for (and I like Visual Studio). :)

Thalhammer:
Because Windows is a pain in the a**....

Sp3EdeR:
hahaha, to each their own! I see both ups and downs for every platform. Specifically for me, MAC/iOS is by far the biggest pain in the a** for development. This is probably not the location to discuss such views however! 🙂

PR #95, opened 2020-07-29
Added EdDSA support

Sp3EdeR:
Added EdDSA support
Fixed build generation for Microsoft Visual Studio
Fixed MSVC build warnings "hides previous local declaration"


Sp3EdeR:
Fixing for OpenSSL 1.0.2g...

tests/CMakeLists.txt
```diff
@@ -33,8 +33,14 @@ target_compile_options(
...
           -Wno-weak-vtables>)
+
+set(dependent_libs "")
+if ( NOT CMAKE_GENERATOR MATCHES "Visual Studio")
+    set(dependent_libs pthread dl)
+endif()
+
 target_link_libraries(jwt-cpp-test PRIVATE jwt-cpp::jwt-cpp GTest::GTest
-                                           GTest::Main pthread dl)
+                                           GTest::Main ${dependent_libs})
```
prince-chrismc, line 43:
```suggestion
                                           GTest::Main $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:pthread ${CMAKE_DL_LIBS}>)
```

tests/CMakeLists.txt
```diff
@@ -33,8 +33,14 @@ target_compile_options(
           -Wno-c++98-compat
           -Wno-global-constructors
           -Wno-weak-vtables>)
+
+set(dependent_libs "")
+if ( NOT CMAKE_GENERATOR MATCHES "Visual Studio")
+    set(dependent_libs pthread dl)
+endif()
+
```
prince-chrismc, line 41:
```suggestion
```

use generator expression below

tests/TokenTest.cpp
```diff
@@ -562,6 +600,52 @@ TEST(TokenTest, VerifyTokenPS256FailNoKey) {
...
 
+#ifndef OPENSSL110
+TEST(TokenTest, VerifyTokenEd25519) {
+	const std::string token = "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXUyJ9.eyJpc3MiOiJhdXRoMCJ9.OujgVcO8xQx5xLcAYWENCRU1SCGH5HcX4MX4o6wU3M4"
+		"DOnKiNmc0O2AnvQlzr-9cgI4QGQzeC6gz_fgLoesADg";
+
+	auto verify = jwt::verify().allow_algorithm(jwt::algorithm::ed25519(ed25519_pub_key, "", "", ""));
+	auto decoded_token = jwt::decode(token);
+
+	verify.verify(decoded_token);
```
prince-chrismc, line 611: please add at least one assertion

```suggestion
	ASSERT_NO_THROW(verify.verify(decoded_token));
```

tests/TokenTest.cpp
```diff
@@ -562,6 +600,52 @@ TEST(TokenTest, VerifyTokenPS256FailNoKey) {
...
+
+TEST(TokenTest, VerifyTokenEd448) {
+	const std::string token = "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXUyJ9.eyJpc3MiOiJhdXRoMCJ9.Aldes9jrXZXxfNjuovqmIZ3r2WF4yVXVr2Q8B8SkAmv"
+		"Bsw_3MHs8HtgKeXbqKFYWpHOCtmZJcH-AWMvoY6FCNdQqbESGTkv58O6tFbXDD_nLejWNAOuvcO2LPMySmkVNQUopmQf_HO62Mug1ngepUDE"
+		"A";
+
+	auto verify = jwt::verify().allow_algorithm(jwt::algorithm::ed448(ed448_pub_key, "", "", ""));
+	auto decoded_token = jwt::decode(token);
+
+	verify.verify(decoded_token);
```
prince-chrismc, line 633:
```suggestion
	ASSERT_NO_THROW(verify.verify(decoded_token));
```

include/jwt-cpp/jwt.h
```diff
@@ -916,6 +921,110 @@ namespace jwt {
 			const size_t signature_length;
 		};
 
+#ifndef OPENSSL110
```
prince-chrismc, line 924: can you share the release notes where this was introduced to openssl?

include/jwt-cpp/jwt.h
```diff
@@ -916,6 +921,110 @@ namespace jwt {
...
+			 * \param signature Signature provided by the jwt
+			 * \param ec Filled with details on error
+			 */
+			void verify(const std::string& data, const std::string& signature, std::error_code& ec) const {
+				ec.clear();
+#ifdef OPENSSL10
+				std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_destroy)> ctx(EVP_MD_CTX_create(), EVP_MD_CTX_destroy);
+#else
+				std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_free)> ctx(EVP_MD_CTX_create(), EVP_MD_CTX_free);
+#endif
```
prince-chrismc, line 995: Since OpenSSL 1.1.1 or greater is required!
```suggestion
				std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_free)> ctx(EVP_MD_CTX_create(), EVP_MD_CTX_free);
```

Thalhammer: Do we really wanna drop support for everything less then 1.1.1 ?
I think removing support for eddsa (assuming its not possible to implement without 1.1.1) in lower versions would be a better solution.

prince-chrismc: thats what's currently proposed, this is just a copy paste error

tests/TokenTest.cpp
Sp3EdeR: Would be okay by me, but no other success-tests have an `ASSERT_NO_THROW` right now, so it would not blend-in. Adding it to all success-tests would probably not fit the PR's topic very well. And gtest catches exceptions and fails the test without an ASSERT_NO_THROW, so it should not really change behaviour.

include/jwt-cpp/jwt.h
Sp3EdeR: https://www.openssl.org/news/openssl-1.1.1-notes.html
> Major changes between OpenSSL 1.1.0i and OpenSSL 1.1.1 [11 Sep 2018]

The one-shot signing and verification functions that EdDSA requires are each new to the 1.1.1 interface, hence the full ifdefs.

include/jwt-cpp/jwt.h
Sp3EdeR: Whoops, missed that one. I did remove the other such place. :)

And yes, the library builds on 1.1.0 without EdDSA right now; and builds with 1.1.1 with EdDSA support.

include/jwt-cpp/jwt.h
```diff
@@ -916,6 +921,106 @@ namespace jwt {
 			const size_t signature_length;
 		};
 
+#ifndef OPENSSL110
+		/**
+		 * \brief Base class for EdDSA family of algorithms
```
prince-chrismc, line 926:
```suggestion
		 * \brief Base class for EdDSA family of algorithms
		 *
		 * These new set of algorithms were introduced with [OpenSLL 1.1.1](https://www.openssl.org/news/openssl-1.1.1-notes.html)
```

include/jwt-cpp/jwt.h
```diff
@@ -916,6 +921,106 @@ namespace jwt {
...
+				: md(md), alg_name(std::move(name))
+			{
+				if (!private_key.empty()) {
+					pkey = helper::load_private_key_from_string(private_key, private_key_password);
+				}
+				else if (!public_key.empty()) {
+					pkey = helper::load_public_key_from_string(public_key, public_key_password);
+				}
+				else
+					throw ecdsa_exception(error::ecdsa_error::load_key_bio_read);
```
prince-chrismc, line 948: should we have a new `eddsa_exception`?

include/jwt-cpp/jwt.h
```diff
@@ -916,6 +921,106 @@ namespace jwt {
...
+					return {};
+				}
+				if (!EVP_DigestSignInit(ctx.get(), nullptr, nullptr, nullptr, pkey.get())) {
+					ec = error::signature_generation_error::signinit_failed;
+					return {};
+				}
+
+				std::string res;
+				size_t len = EVP_PKEY_size(pkey.get());
+				res.resize(len);
```
prince-chrismc, line 970:
```suggestion
				size_t len = EVP_PKEY_size(pkey.get());
				std::string res(len, '\0');
```

prince-chrismc:
Also it looks like there's the new OpenSSL tests which needs to be added for the algorithm

include/jwt-cpp/jwt.h
Sp3EdeR: EdDSA is also a kind of elliptic curve DSA. I didn't think it necessary to specialize the error messages further.

prince-chrismc:
Ill try to add them this evening =)

Sp3EdeR:
I've only updated the algorithm listing, including version information rather than a lengthy talk about the new feature. I decided so because the readme does not include a version history and the lengthy chat would kind of been a version history thing.

But here's a question. Shouldn't the library get a version history?

include/jwt-cpp/jwt.h
prince-chrismc: it would only be used once :thinking: I agree

prince-chrismc:
I like your approach. It's version though the releases (tags in git) which are more then enough

:+1:

prince-chrismc:
> Coverage increased (+0.08%) to 96.452%

Success tastes good. :cake:

tests/OpenSSLErrorTest.cpp
```diff
@@ -645,5 +708,90 @@ TEST(OpenSSLErrorTest, PS256VerifyErrorCode) {
...
+        { &fail_EVP_DigestVerify, 1, jwt::error::signature_verification_error::verifyfinal_failed }
+    };
+
+    run_multitest(mapping, [&alg, &signature](std::error_code& ec){
+        alg.verify("testdata", signature, ec);
+    });
+}
+
 #endif
 #endif
```
prince-chrismc, line 797:
```suggestion
#endif

```

tests/OpenSSLErrorTest.cpp
```diff
@@ -507,10 +575,6 @@ TEST(OpenSSLErrorTest, RS256VerifyErrorCode) {
     });
 }
 
-TEST(OpenSSLErrorTest, ECDSAKeyReference) {
```
Thalhammer, line 510: Is there a reason why you remove the reference test for ECDSAKey ?

prince-chrismc: This is already done the token tests and it's not  an "openssl error" test, plus it's the only key reference test

PR #102, opened 2020-10-16
Remove toolchain include

zxey:
When you specify toolchain file, cmake makes the variables from it available to all CMakeLists, that is cmake includes this file for you if you specifiy -DCMAKE_TOOLCHAIN_FILE.

This also breaks crosscompilation for me when I use this library, when removed everything works fine.

What is the reasoning behind this? Am I missing something?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,5 @@
 cmake_minimum_required(VERSION 3.8)
 
-if(CMAKE_TOOLCHAIN_FILE)
-  include(${CMAKE_TOOLCHAIN_FILE})
-endif()
-
 option(BUILD_TESTS "Configure CMake to build tests (or not)" ON)
 option(BUILD_EXAMPLES "Configure CMake to build examples (or not)" ON)
 option(COVERAGE "Enable code coverage testing" OFF)


prince-chrismc:
I had added it to allow easy support for vcpkg and conan integration, what toolchain are you specifying?

Thalhammer:
I think his point is that you don't actually need to include it manually as CMake will include it itself **before** everything else in the CMakeLists and he is actually right about that. Including it manually here probably screws things up as when crosscompiling you usually discard all variables (compiler & flags) and start from scratch to make sure you have a clean environment. I dont see any reason why we should need to include it.

zxey:
I specify custom toolchain to crosscompile for arm, looks similar to [this](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-linux). I am not sure about conan, but vpckg definitely works without the include.

prince-chrismc:
Wow thanks! I have done little cross compilation with CMake I was not familiar with this behavior... such is life when you copy a bad example.

Thank you for the contribution 🥇

prince-chrismc (approved):

PR #106, opened 2020-11-03
Add helper function to convert cert from base64DER to PEM

jbajwa:
This is particularly useful when dealing with JWKs x5c claim which is encoded as base64 DER. With this helper function the client can convert x5c claim to PEM, which makes it easy to interact with this library.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -397,6 +397,41 @@ namespace jwt {
 		}
 
 		/**
+		 * \brief Convert the cert provided as base64 DER to PEM. This useful when
+		 * using with JWKs as x5c claim is encoded as base64 DER. More info 
+		 * (here)[https://tools.ietf.org/html/rfc7517#section-4.7]
+		 * 
+		 * \param cert_base64_der_str String containing the certificate encoded as base64 DER
+		 * \throws rsa_exception if an error occurred
+		 */
+		inline
+		std::string convert_base64_der_to_pem(const std::string& cert_base64_der_str) {
+
+			auto decodedStr = base::decode<alphabet::base64>(base::pad<alphabet::base64>(cert_base64_der_str));
+			auto c_str = reinterpret_cast<const unsigned char *> (decodedStr.c_str());
+				
+			//
+			// d2i -> DER to internal x509 struct
+			//
+			std::unique_ptr<X509, decltype(&X509_free)> cert(d2i_X509(NULL, & c_str, decodedStr.size()), X509_free);
+			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new(BIO_s_mem()), BIO_free_all);
+			if(!cert || !certbio) {
+				throw rsa_exception(error::rsa_error::create_mem_bio_failed);
+			}
+
+			PEM_write_bio_X509(certbio.get(), cert.get());
+
+			char* ptr = nullptr;
+			auto len = BIO_get_mem_data(certbio.get(), &ptr);
+			if(len <= 0 || ptr == nullptr) {
+				throw rsa_exception(error::rsa_error::convert_to_pem_failed);
+			}
+
+			std::string cert_pem(ptr, len);
+			return cert_pem;
+		}
+
+		/**
 		 * \brief Load a public key from a string.
 		 * 
 		 * The string should contain a pem encoded certificate or public key
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -427,13 +427,13 @@ namespace jwt {
 			}
 
 			char* ptr = nullptr;
-			const uint64_t len = BIO_get_mem_data(certbio.get(), &ptr);
+			const auto len = BIO_get_mem_data(certbio.get(), &ptr);
 			if(len <= 0 || ptr == nullptr) {
 				ec = error::rsa_error::convert_to_pem_failed;
 				return {};
 			}
 
-			return std::string {ptr, len};
+			return std::string {ptr, static_cast<size_t>(len)};
 		}
 
 		/**
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -397,26 +397,27 @@ namespace jwt {
 		}
 
 		/**
-		 * \brief Convert the cert provided as base64 DER to PEM. This useful when
-		 * using with JWKs as x5c claim is encoded as base64 DER. More info 
+		 * \brief Convert the certificate provided as base64 DER to PEM.
+		 *
+		 * This is useful when using with JWKs as x5c claim is encoded as base64 DER. More info
 		 * (here)[https://tools.ietf.org/html/rfc7517#section-4.7]
-		 * 
+		 *
 		 * \param cert_base64_der_str String containing the certificate encoded as base64 DER
-		 * \throws rsa_exception if an error occurred
+		 * \param ec		error_code for error_detection (gets cleared if no error occures)
 		 */
 		inline
-		std::string convert_base64_der_to_pem(const std::string& cert_base64_der_str) {
-
+		std::string convert_base64_der_to_pem(const std::string& cert_base64_der_str, std::error_code& ec) {
+			ec.clear();
 			auto decodedStr = base::decode<alphabet::base64>(base::pad<alphabet::base64>(cert_base64_der_str));
 			auto c_str = reinterpret_cast<const unsigned char *> (decodedStr.c_str());
-				
 			//
 			// d2i -> DER to internal x509 struct
 			//
 			std::unique_ptr<X509, decltype(&X509_free)> cert(d2i_X509(NULL, & c_str, decodedStr.size()), X509_free);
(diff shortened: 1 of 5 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -396,6 +396,41 @@ namespace jwt {
 			return res;
 		}
 
+		/**
+		 * \brief Convert the cert provided as base64 DER to PEM. This useful when
+		 * using with JWKs as x5c claim is encoded as base64 DER. More info 
+		 * (here)[https://tools.ietf.org/html/rfc7517#section-4.7]
```
prince-chrismc, line 402:
```suggestion
		 * \brief Convert a certificate provided as base64 DER to PEM.
		 *
		 * This useful when using with JWKs as x5c claim is encoded as base64 DER.
		 * More info (here)[https://tools.ietf.org/html/rfc7517#section-4.7]
```

tests/ClaimTest.cpp
```diff
@@ -59,7 +59,7 @@ TEST(ClaimTest, AudienceAsSet) {
 TEST(ClaimTest, SetAudienceAsSet) {
 	auto token = jwt::create()
 		.set_type("JWT")
-		.set_audience({picojson::value("test"), picojson::value("test2")})
+		.set_audience({{picojson::value("test"), picojson::value("test2")}})
```
prince-chrismc, line 62: What is the purpose of this change?

prince-chrismc (commented):
Thank you for the contribution 🥇 

I really like the idea of adding small helpers like this one, it makes the library more attractive. 🚀 

In order for this to get merged it, there's small speed dump, the library offers an `std::error_code` form for all the functions.

Example: https://github.com/Thalhammer/jwt-cpp/blob/012a2fce5104cee65d60fd9ddc33e8e60b3ff56d/include/jwt-cpp/jwt.h#L392-L397

It's important to have good test coverage as well, would it be possible for you to add some error scenario test code? 🙏 

Thanks for taking the time to open this PR 👍

tests/ClaimTest.cpp
jbajwa: I was getting the following error when compiling with (clang-1100.0.33.17)

`error: call to member function 'set_audience' is ambiguous
                .set_audience({picojson::value("test"), picojson::value("test2")})
                ~^~~~~~~~~~~~
.../jwt-cpp/include/jwt-cpp/jwt.h:2263:12: note: candidate function
                builder& set_audience(typename json_traits::array_type a) { return set_payload_claim("aud", typename json_traits::value_type(a)); }
                         ^
.../jwt-cpp/include/jwt-cpp/jwt.h:2269:12: note: candidate function
                builder& set_audience(typename json_traits::string_type aud) { return set_payload_claim("aud", typename json_traits::value_type(aud)); }`

maybe a clang thing, I don't see this on gcc 9.2.0.

prince-chrismc: Hmmm interesting!

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +396,60 @@ namespace jwt {
...
+			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new(BIO_s_mem()), BIO_free_all);
+			if(!cert || !certbio) {
+				ec = error::rsa_error::create_mem_bio_failed;
+				return {};
+			}
+
+			PEM_write_bio_X509(certbio.get(), cert.get());
+
+			char* ptr = nullptr;
+			auto len = BIO_get_mem_data(certbio.get(), &ptr);
```
prince-chrismc, line 426:
```suggestion
			const auto len = BIO_get_mem_data(certbio.get(), &ptr);
```

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +396,60 @@ namespace jwt {
...
+			// d2i -> DER to internal x509 struct
+			//
+			std::unique_ptr<X509, decltype(&X509_free)> cert(d2i_X509(NULL, & c_str, decodedStr.size()), X509_free);
+			std::unique_ptr<BIO, decltype(&BIO_free_all)> certbio(BIO_new(BIO_s_mem()), BIO_free_all);
+			if(!cert || !certbio) {
+				ec = error::rsa_error::create_mem_bio_failed;
+				return {};
+			}
+
+			PEM_write_bio_X509(certbio.get(), cert.get());
```
prince-chrismc, line 423: Do we need handle the error return here?

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +396,60 @@ namespace jwt {
...
+		 */
+		inline
+		std::string convert_base64_der_to_pem(const std::string& cert_base64_der_str, std::error_code& ec) {
+			ec.clear();
+			auto decodedStr = base::decode<alphabet::base64>(base::pad<alphabet::base64>(cert_base64_der_str));
+			auto c_str = reinterpret_cast<const unsigned char *> (decodedStr.c_str());
+			//
+			// d2i -> DER to internal x509 struct
+			//
+			std::unique_ptr<X509, decltype(&X509_free)> cert(d2i_X509(NULL, & c_str, decodedStr.size()), X509_free);
```
prince-chrismc, line 416: I also checked, https://github.com/openssl/openssl/blame/e84b663a933eff8834d66233ac0d6882ed918b82/doc/crypto/d2i_X509.pod https://www.openssl.org/docs/man1.0.2/man3/d2i_X509.html

Should be safe it's in the 1.0.X

```suggestion
			std::unique_ptr<X509, decltype(&X509_free)> cert(d2i_X509(NULL, & c_str, decodedStr.size()), X509_free);
```

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +396,60 @@ namespace jwt {
...
+		 * This is useful when using with JWKs as x5c claim is encoded as base64 DER. More info
+		 * (here)[https://tools.ietf.org/html/rfc7517#section-4.7]
+		 *
+		 * \param cert_base64_der_str String containing the certificate encoded as base64 DER
+		 * \param ec		error_code for error_detection (gets cleared if no error occures)
+		 */
+		inline
+		std::string convert_base64_der_to_pem(const std::string& cert_base64_der_str, std::error_code& ec) {
+			ec.clear();
+			auto decodedStr = base::decode<alphabet::base64>(base::pad<alphabet::base64>(cert_base64_der_str));
```
prince-chrismc, line 411:
```suggestion
			const auto decodedStr = base::decode<alphabet::base64>(base::pad<alphabet::base64>(cert_base64_der_str));
```

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +396,60 @@ namespace jwt {
...
+
+			char* ptr = nullptr;
+			auto len = BIO_get_mem_data(certbio.get(), &ptr);
+			if(len <= 0 || ptr == nullptr) {
+				ec = error::rsa_error::convert_to_pem_failed;
+				return {};
+			}
+
+			std::string cert_pem(ptr, len);
+			return cert_pem;
```
prince-chrismc, line 433:
```suggestion
			return std::string {ptr, len};
```

prince-chrismc (commented):
Few more editorial changes but LGTM 🎉

include/jwt-cpp/jwt.h
jbajwa: yup, I think we should.

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +398,61 @@ namespace jwt {
...
+			}
+
+			if(!PEM_write_bio_X509(certbio.get(), cert.get()))
+			{
+				ec = error::rsa_error::write_cert_failed;
+				return {};
+			}
+
+			char* ptr = nullptr;
+			const uint64_t len = BIO_get_mem_data(certbio.get(), &ptr);
```
jbajwa, line 430: auto was deducing len as int but string expects size_t (uint). Was getting `warning: narrowing conversion` so decided to be explicit with the type here.

prince-chrismc (approved):
LGTM 🚀

include/jwt-cpp/jwt.h
Thalhammer: Why not use size_t instead ?
Although size_t probably is uint64_t on any sane platform, its not guranteed and I don't see any reason to use uint64_t.
Also using size_t sounds better imho since it indicates a size.

include/jwt-cpp/jwt.h
```diff
@@ -396,6 +398,61 @@ namespace jwt {
...
+			if(len <= 0 || ptr == nullptr) {
```
Thalhammer, line 431: This check is wrong as len is unsigned and thus can never be less than 0.
Actually I am a bit surprised the compiler does not warn about this.

Thalhammer (changes requested):
See comments

include/jwt-cpp/jwt.h
jbajwa: ah, good catch, so I suppose I can't use an unsigned as `BIO_get_mem_data` could return a negative value in case of an error. I'll cast len below in the string constructor.

Thalhammer (approved):

Issue #94, opened 2020-07-29
Question about JOSE


mohsenomidi:
Dear @Thalhammer 

Thanks for this useful library and recent changes are wonderful, we can use the json library like simdjson or rapidjson for performance point of view

Since you develop this jwt library, do you have any plan to imitate the library for JOSE completely, I mean jwe, jws, jwk, etc?

Thalhammer:
Hi,
While I absolutely think those might give a great addition without to much added stuff, I honestly don't have the time right now to read through and implement those.

However @prince-chrismc did a lot of contributions lately and @Sp3EdeR works on implementing EdDSA support right now, so going forward I am pretty sure we will implement it eventually. It's just that theres no timeline for it.
If you need it soon and wanna implement it yourself I would ofc happily merge a PR if it matches the overall style of the library.

kleinmrk:
@Thalhammer @prince-chrismc what is your take on loading keys from JWK? I [experimented with it a bit](https://github.com/kleinmrk/jwt-cpp/blob/support-loading-jwk-from-json/tests/JwkTest.cpp#L50-L54) and I can also contribute a minimal working feature but it feels like the interface is not optimal and I would like to avoid putting effort into something that will be dismissed as "clumsy to use".

Namely, it looks like there is missing a class that represents keys. @Thalhammer already mentioned something similar [here](https://github.com/Thalhammer/jwt-cpp/pull/146#issuecomment-871220066). Keys are currently loaded directly from certificates/PEM encoded public keys and stored in `jwt::algorithm` [structs](https://github.com/Thalhammer/jwt-cpp/blob/9d1a010fea3ed3293112edec9745dc435a85a849/tests/TokenTest.cpp#L261). Does it make sense to introduce a class that encapsulates a key that one can pass to `jwt::verifier` and `jwt::builder`? The code could look somewhat like this:

```c++
jwt::jwk key = jwk::build_from_certificate("-----BEGIN CERTIFICATE...")
auto verify = jwt::verify().allow_algorithm(jwt::algorithm::rs256(key));
verify.verify(jwt::decode(token));
```
and in future one could implement reading JWK from JSON
```c++
jwt::jwk key = jwk::build_from_json(R"({"kty" : "oct", "k" : "FdFYFzERwC2uCBB46pZQi4GG85LujR8obt-KWRBICVQ"})")
auto verify = jwt::verify().allow_algorithm(jwt::algorithm::hs256(key));
[...]

PR #112, opened 2020-12-02
Add jwks parsing (abandoned)

dtiukalov:
add jwks implementation to simple access to x5c to verify jwt:

```cpp
auto decoded_jwt = jwt::decode(token);
auto jwks = jwt::parse_jwks(jwks_json_str);
if (jwks.has_x5c_id()) {
   auto issuer = decoded_jwt.get_issuer();
   auto x5c = jwks.get_x5c();
   if (jwks.get_key_id() == decoded_jwt.get_key_id()) {
      if (!x5c.empty() && !issuer.empty()) {
         auto verifier = jwt::verify()
            .allow_algorithm(jwt::algorithm::rs256(jwt::helper::convert_base64_der_to_pem(x5c), "", "", ""))
            .with_issuer(issuer)
            .leeway(60U); 
         std::error_code err;
         verifier.verify(decoded_jwt, err);
         jwt::error::throw_if_error(err);
     }
}
   else {
      std::cerr << "Key ids are not matched" << std::endl;
   }
}
```


prince-chrismc:
Wow! 🏆 super exciting to see this added, thank you so much for your contribution ❤️

This would be a start towards #94 which I think a lot of people would benefit from!

There's a few key details missing in order for this to be merged

- Test code
   - Something similar to [this test](https://github.com/Thalhammer/jwt-cpp/blob/master/tests/ClaimTest.cpp#L35)
   - Feel free to add a new file
- No base64 (required by the helper you are using)
   - You'll need to add a overload like this https://github.com/Thalhammer/jwt-cpp/blob/ca110ade5bf70920175bd6679c39e2c0f79239d7/include/jwt-cpp/jwt.h#L420-L421
- Example
  - I think the code you wrote above is excellent, just stick it in the examples!

I very happy to see this PR get merged, let me know if you need more help! 🚀

example/jwks-verify.cpp
```diff
@@ -0,0 +1,52 @@
...
+	auto jwks = jwkskeys.get_jwks(decoded_jwt.get_key_id());
+
+  auto issuer = decoded_jwt.get_issuer();
+  auto x5c = jwks.get_x5c();
+
+  if (!x5c.empty() && !issuer.empty()) {
+    auto verifier = jwt::verify()
+      .allow_algorithm(jwt::algorithm::rs256(jwt::helper::convert_base64_der_to_pem(x5c), "", "", ""))
+      .with_issuer(issuer)
+      .leeway(10000000000UL); // value in seconds, add some to compensate timeout
```
prince-chrismc, line 44: Perhaps, a simple token with no expiration is possible?

example/jwks-verify.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    std::error_code err;
+    verifier.verify(decoded_jwt, err);
+
+    jwt::error::throw_if_error(err);
```
prince-chrismc, line 49:
```suggestion
    verifier.verify(decoded_jwt);
```

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2668,162 @@ namespace jwt {
 			}
 		}
 	};
+	
+	template<typename json_traits>
+	class jwks {
```
prince-chrismc, line 2673: This seems to more closely match a `JWK` from https://tools.ietf.org/html/rfc7517 than the `JWK Set` from https://tools.ietf.org/html/rfc7517#section-5

I think the naming here and for `jwks_keys` are a little confusing for me.

Perhaps this could be called a `jwk` since it's one single key descriptor, and the following would be a `jwks`

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2668,162 @@ namespace jwt {
...
+		using basic_claim_t = basic_claim<json_traits>;
+		template<typename> friend class jwks_keys;
+
+	public:
+		JWT_CLAIM_EXPLICIT jwks(const typename json_traits::string_type& jwks_token) {
+			auto parse_claims = [](const typename json_traits::string_type& str) {
```
prince-chrismc, line 2679: this block seems to be copy pasted from the original

📓 for myself to look into a small refactor!

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2668,162 @@ namespace jwt {
 			}
 		}
 	};
+	
+	template<typename json_traits>
+	class jwks {
+		using basic_claim_t = basic_claim<json_traits>;
+		template<typename> friend class jwks_keys;
```
prince-chrismc, line 2675: I am not convinced this is required

You could simply make the `protected` ctor public and remove this friend association

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2668,162 @@ namespace jwt {
...
+		 * Get jwks claim
+		 * \return Requested claim
+		 * \throw std::runtime_error If claim was not present
+		 */
+		basic_claim_t get_jwks_claim(const typename json_traits::string_type& name) const {
+			if (!has_jwks_claim(name)) {
+				typename json_traits::string_type error_msg = "claim \"";
+				error_msg += name;
+				error_msg += "\" not found";
+				throw std::runtime_error(error_msg.c_str());
```
prince-chrismc, line 2754: Since all of of them should be changed in the future...

```suggestion
				throw std::runtime_error("claim not found");
```

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2668,162 @@ namespace jwt {
...
+			auto parse_claims = [](const typename json_traits::string_type& str) {
+				std::unordered_map<typename json_traits::string_type, jwks_t> res;
+				typename json_traits::value_type val;
+				if (!json_traits::parse(val, str))
+					throw std::runtime_error("Invalid json");
+
+				for (const auto& k : json_traits::as_object(val)) {
+					for (const typename json_traits::value_type& item : json_traits::as_array(k.second)) {
+						jwks_t jwks_entry(item);
+						res.emplace(jwks_entry.get_key_id(), std::move(jwks_entry));
```
prince-chrismc, line 2789: What if there is no KID in a JWK? there's still a lot of auth server which do not follow this best practice

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2668,162 @@ namespace jwt {
...
+				jwks_claims.emplace(e.first, basic_claim_t(e.second));
+			}
+		}
+
+	private:
+		std::unordered_map<typename json_traits::string_type, basic_claim_t> jwks_claims;
+	};
+
+	template<typename json_traits>
+	class jwks_keys {
```
prince-chrismc, line 2773: I wonder if this "parser" could be a static method that returns a vector, this was the implementation is not limited on KIDs being used in all the keys

Thoughts?

dtiukalov: to_vector() is added

include/jwt-cpp/jwt.h
dtiukalov: yes, and it costs an extra move, good point to refactor.

include/jwt-cpp/jwt.h
```diff
@@ -2699,130 +2707,133 @@ namespace jwt {
...
+		typename json_traits::string_type get_key_id() const { return get_jwk_claim("kid").as_string(); }
 
 		/**
 		 * Get x5c claim
 		 * \return x5c as string
 		 * \throw std::runtime_error If claim was not present
 		 * \throw std::bad_cast Claim was present but not a string (Should not happen in a valid token)
 		 */
-		typename json_traits::string_type get_x5c() const { return json_traits::as_string(get_jwks_claim("x5c").as_array().front()); } //do not use serialize() instead
+		typename json_traits::string_type get_x5c() const { return json_traits::as_string(get_jwk_claim("x5c").as_array().front()); } //do not use serialize() instead
```
prince-chrismc, line 2726: I think the consumer will/may need the array in larger systems as per [the rfc 4.7](https://tools.ietf.org/html/rfc7517#section-4.7)

> The PKIX certificate containing the key value MUST be the first
   certificate.  This MAY be followed by additional certificates, with
   each subsequent certificate being the one used to certify the
   previous one.

What do you think about having two functions?

```cpp
typename json_traits::array_type get_x5c() const;
typename json_traits::string_type get_x5c_key_value() const;
```

Secondly, `front` might be a little dangerous! [Calling front on an empty container is undefined](https://en.cppreference.com/w/cpp/container/vector/front)

please throw if empty 💥

include/jwt-cpp/jwt.h
```diff
@@ -2699,130 +2707,133 @@ namespace jwt {
...
 		 */
-		jwks_t get_jwks(const typename json_traits::string_type& key_id) const {
-			if (!has_jwks(key_id)) {
-				typename json_traits::string_type error_msg = "jwks with key id \"";
+		jwk_t get_jwk(const typename json_traits::string_type& key_id) const {
+			if (!has_jwk(key_id)) {
+				typename json_traits::string_type error_msg = "jwk with key id \"";
 				error_msg += key_id;
 				error_msg += "\" not found";
 				throw std::runtime_error(error_msg.c_str());
```
prince-chrismc: the caller knows the kid they passed, no need to repeat it

```suggestion
				throw std::runtime_error("jwk with matching key id not found");
```

include/jwt-cpp/jwt.h
```diff
@@ -2699,130 +2707,133 @@ namespace jwt {
...
+			v_jwks_claims.reserve(jwks_claims.size());
+			for (const auto& element : jwks_claims)
+   				v_jwks_claims.push_back(element.second);
+			return v_jwks_claims;
+		}
 
 	private:
 
-		std::unordered_map<typename json_traits::string_type, jwks_t> jwks_keys_claims;
+		std::unordered_map<typename json_traits::string_type, jwk_t> jwks_claims;
```
prince-chrismc, line 2836: Only because i've implemented this and used a few different auth servers... I am really unsure about this being a map...

What if there are two keys neither have KIDs? One would be lost, most likely the new key that is starting to issue tokens.

prince-chrismc (commented):
The changes look really good! 👑 

There's a few more details about `x5c` handling I would appreciate your thoughts on.

Lastly, I need to get my note book and review them, I may come back with a suggestion to make the `jwks` more generic... there's a lot of variety with the content and I want to make sure. Hopefully you can understand!

Keep up the awesome work 💪

include/jwt-cpp/jwt.h
```diff
@@ -2668,6 +2671,170 @@ namespace jwt {
...
+					size_t id = 0;
+					for (const typename json_traits::value_type& item : json_traits::as_array(k.second)) {
+						jwk_t jwk_entry(item);
+						if (jwk_entry.has_key_id())
+							res.emplace(jwk_entry.get_key_id(), std::move(jwk_entry));
+						else
+							res.emplace(std::to_string(id), std::move(jwk_entry));
+						++id;
+					}
+				}
```
prince-chrismc, line 2794: Yep, so I would really like if we could leave the KID outside of the JWKS logic because it's important to support workflows that do not required them. 

I will come back with some changes against your branch, I think the biggest limitation is the parsing or "map of claims"

I also see we don't check `"keys"` explicitly is the JWKS which would open use to a vulnerability

prince-chrismc:
@dtiukalov please consider https://github.com/dtiukalov/jwt-cpp/pull/1

rasmuspeders1:
This looks great! Basically this is exactly what I was looking for. A single nice c++ library to handle jwt auth including jwk handling.
Looking forward to trying it out when it is ready.

prince-chrismc:
> Looking forward to trying it out when it is ready.

Sadly it seems OP has stopped working on this, it would be greatly appreciated if you could pick up and work on this!

Lots of work has been done just needs a little push to get it over

rasmuspeders1:
> > Looking forward to trying it out when it is ready.
> 
> Sadly it seems OP has stopped working on this, it would be greatly appreciated if you could pick up and work on this!
> 
> Lots of work has been done just needs a little push to get it over

That is too bad. 
I am not sure I will have the time in the near future nor possess the necessary skill. I will keep it in mind if I find some time.

Tchinmai7:
@prince-chrismc what's left in this PR? maybe i can take a shot at this

Thalhammer:
It will probably also need a pretty significant update to the current version.

Tchinmai7:
Does it make sense to start a new PR with OP's changes + @prince-chrismc 's suggestions to it, targetting the latest version? I think looking at the latest diff might help us see what's left

PR #114, opened 2020-12-06
Fix exported include path

aboseley:
GNUInstallDirs is used before it is defined

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ endif()
 if(JWT_DISABLE_PICOJSON)
   target_compile_definitions(jwt-cpp INTERFACE JWT_DISABLE_PICOJSON)
 endif()
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
 target_include_directories(jwt-cpp INTERFACE $<BUILD_INTERFACE:${JWT_INCLUDE_PATH}>
                                              $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 target_link_libraries(jwt-cpp INTERFACE OpenSSL::SSL OpenSSL::Crypto)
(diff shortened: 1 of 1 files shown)


prince-chrismc:
Hmmm, I was under the impression, the `include` happened during configure and the generator expression at generate... I haven't seen any issue with this.

Under what conditions did you encounter this? I would like to add a test for it! Make sure it does not break in the future

prince-chrismc (approved):

aboseley:
Before this change the exported and installed file "cmake/jwt-cpp/jwt-cpp-targets.cmake" . Was missing the extra line, to tell consuming project what the include paths should be added when using the jwt-library
` INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"`

So when I tried to used the install package config file via,  "FindPackage( jwt-cpp)", then added it as a dependency of my application via "target_link_libraries" the include paths weren't getting set correctly.

This only shows when you  install the jwt-cpp package to a non-standard directly where this becomes important.  In my particular case I'm install jwt-cpp to "/opt/jwt-cpp", so I'm relying on the imported target include paths.

On that note, the naming of the library is somewhat unfortunate.   With cmake an easy way to direct find-package to a non standard install is via <packagename>_ROOT environment variable but the 'dash' in the package-name make's this impossible in linux/Bash.

prince-chrismc:
> This only shows when you install the jwt-cpp package to a non-standard directory 

This is something I was expecting as I tend to avoid this. I'll look into adding a test to make sure this use case works in the future

🤯 There is one too many ways for `find_package` to work. 
https://cmake.org/cmake/help/latest/envvar/PackageName_ROOT.html I was not aware of this feature.
🤣 I taught I was pretty familiar with CMake but there's always new (or old) ways of doing things.

PR #116, opened 2020-12-24
adding test for non-default cmake install prefix

prince-chrismc:
This is to follow up from Thalhammer#114 (comment)

Related CMake docs: https://cmake.org/cmake/help/v3.8/variable/CMAKE_INSTALL_PREFIX.html

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -47,7 +47,7 @@ jobs:
       run: |
         mkdir build
         cd build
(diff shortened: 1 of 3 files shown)


.github/workflows/cmake.yml
```diff
@@ -36,6 +36,26 @@ jobs:
         cmake . -DCMAKE_PREFIX_PATH=/usr/local/cmake -DTEST:STRING="defaults-enabled"
         cmake --build .
 
+  costume-install:
```
prince-chrismc, line 39: fix bad english

```suggestion
  custom-install:
```

PR #117, opened 2020-12-31
:control_knobs: Expand support to include non STL container based JSON implementations

prince-chrismc:
My goal was to validate the traits implementation we worked on against a library which took a less orthodox approach. I chose [`jsoncons`](https://github.com/danielaparker/jsoncons) for its rich feature set and support for wide strings.

Unlike `picojson` and `nlohmann_json`, the internal implementation of `jsoncons` is based around custom objects which have a stripped down minimalist API.

### Major Changes
- Refactor of the claims parsing.
   - new `map_of_claim`
   - replaced `std::unordered_map<string_type, basic_claim>` with `object_type` on the ingest side.
   - this has a nice advantage of only paying for the transformation when it is required 👍 for reducing overhead
- More traits checking

🗒️ There is no API break so it should be a freebie for consumers to upgrade!


include/jwt-cpp/jwt.h
```diff
@@ -2219,15 +2348,15 @@ namespace jwt {
 		 * Get all payload claims
 		 * \return map of claims
 		 */
-		std::unordered_map<typename json_traits::string_type, basic_claim<json_traits>> get_payload_claims() const {
-			return this->payload_claims;
+		std::unordered_map<typename json_traits::string_type, basic_claim_t> get_payload_claims() const {
```
Thalhammer, line 2351: Returning map_of_claims would probably look better and cheaper (e.g. by returning a const reference to the member).
However it would be a breaking api change.
We already do have a bunch of breaking changes for 0.4.0 => 0.5.0 so everyone thats not living on master has to check their code anyway and everyone living on master should IMHO expect breaking changes at any time.

But its up to you. Maybe a good middleground would be adding a `get_payload_claims_map()` and deprecating this one. Not having to convert to unordered_map would also remove the std::pair requirement noted on line 2045.

Thalhammer (approved):
Looks good, but I added some minor ideas.

include/jwt-cpp/jwt.h
prince-chrismc: There's actually no API breaks, looking at  https://github.com/Thalhammer/jwt-cpp/compare/v0.4.0...master

There are only this change in examples https://github.com/Thalhammer/jwt-cpp/compare/v0.4.0...master#diff-c5ace5b1f9f956b086a7e09a83c2fe4910841987ddc1dc08887a1f5356a83ad1L9 which is from added stream operators

This might be an API break https://github.com/Thalhammer/jwt-cpp/compare/v0.4.0...master#diff-d2b23c35f69c8a9a09ea6dae014a1e5b4b888f6a0d5a6450a8662b7bcf3629d3R83 🤔 

The rest of the changes in examples and tests are new code.

***

I would love to remove these functions! They are only required for the "printing claims example" and provide no functionality for the user... less code less bugs :bug: ?

Right now the `std::pair` is optional, but required for the example.

tests/NlohmannTest.cpp
```diff
@@ -10,29 +10,29 @@ struct nlohmann_traits {
...
+	using integer_type = json::number_integer_t;
+	using boolean_type = json::boolean_t;
 	
 	static jwt::json::type get_type(const json &val) {
 		using jwt::json::type;
 
 		if (val.type() == json::value_t::boolean)
 			return type::boolean;
-		else if (val.type() == json::value_t::number_integer)
+		if (val.type() == json::value_t::number_integer)
```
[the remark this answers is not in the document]
prince-chrismc: > if a value is missing

We 💯 are 
https://github.com/nlohmann/json/blob/68c36963826671d3f3ba157222430109ef932bac/single_include/nlohmann/json.hpp#L3463-L3475

```cpp
enum class value_t : std::uint8_t
{
    null,             ///< null value
    object,           ///< object (unordered set of name/value pairs)
    array,            ///< array (ordered collection of values)
    string,           ///< string value
    boolean,          ///< boolean value
    number_integer,   ///< number value (signed integer)
    number_unsigned,  ///< number value (unsigned integer)
    number_float,     ///< number value (floating-point)
    binary,           ///< binary array (ordered collection of bytes)
    discarded         ///< discarded by the parser callback function
};
```

This is a fairly complete list of JSON types, and we do not support (or better do not require) the `null` or `binary` types.

I'm indifferent.

PR #119, opened 2021-01-07
conan example added

mbaykara:
example adopted with Conan based on CMake


prince-chrismc (commented):
How would you feel about only using one cmake file? I think that's better for show how it works... you script as a few extras I was not expecting.

My personal preference is for the cmake_find_package[_multi] generators. I think it's nice to show the similarities.

mbaykara:
I've changed accordingly and removed the second CMake file

conan-example/CMakeLists.txt
```diff
@@ -1,15 +1,7 @@
 cmake_minimum_required(VERSION 3.16)
```
prince-chrismc, line 1:
```suggestion
cmake_minimum_required(VERSION 3.8)
```

prince-chrismc (changes requested):
Super close!

mbaykara:
it should be done now

prince-chrismc (approved):

PR #142, opened 2021-05-10
Changed the JWT_CMAKE_FILES_INSTALL_DIR such that it is found on both…

sweco-nlahoo:
… Windows and Unix and that it may be configured to something entirely different

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -54,7 +54,7 @@ jobs:
       - name: test
         run: |
           cd tests/cmake
-          cmake . -DCMAKE_PREFIX_PATH=/opt/jwt-cpp/cmake -DTEST:STRING="defaults-enabled"
+          cmake . -DCMAKE_PREFIX_PATH="/opt/jwt-cpp" -DTEST:STRING="defaults-enabled"
           cmake --build .
 
   custom-install-win:
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -34,7 +34,7 @@ jobs:
       - name: test
         run: |
           cd tests/cmake
-          cmake . -DCMAKE_PREFIX_PATH=/usr/local/cmake -DTEST:STRING="defaults-enabled"
+          cmake . -DTEST:STRING="defaults-enabled"
           cmake --build .
 
   custom-install-linux:
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -37,7 +37,7 @@ jobs:
           cmake . -DCMAKE_PREFIX_PATH=/usr/local/cmake -DTEST:STRING="defaults-enabled"
           cmake --build .
 
-  custom-install:
+  custom-install-linux:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/cmake.yml
```diff
@@ -57,6 +57,26 @@ jobs:
           cmake . -DCMAKE_PREFIX_PATH=/opt/jwt-cpp/cmake -DTEST:STRING="defaults-enabled"
```
prince-chrismc, line 57: Assuming we change the path this should now work

```suggestion
          cmake . -DCMAKE_PREFIX_PATH="/opt/jwt-cpp" -DTEST:STRING="defaults-enabled"
```

prince-chrismc (commented):
Are you using the default install paths? Something like

```sh
jwt-cpp $ cmake .
jwt-cpp $ cmake --install .
my_project $ cmake . # find_package(jwt-cpp REQUIRED)
```

Or are you doing something like the tests by explicitly setting the "install prefix"?

I am starting to suspect https://github.com/Thalhammer/jwt-cpp/blob/0d10b631ea0e1cf5f91b666aaa3e48d71d89fa5e/.github/workflows/cmake.yml#L37 was just a patch on my part since with your change it still works 🤔

prince-chrismc:
Thanks for you patients 🤗 I just want to make sure the fix is good

sweco-nlahoo:
> Are you using the default install paths? Something like
> 
> ```shell
> jwt-cpp $ cmake .
> jwt-cpp $ cmake --install .
> my_project $ cmake . # find_package(jwt-cpp REQUIRED)
> ```
> 
> Or are you doing something like the tests by explicitly setting the "install prefix"?
> 
> I am starting to suspect
> 
> https://github.com/Thalhammer/jwt-cpp/blob/0d10b631ea0e1cf5f91b666aaa3e48d71d89fa5e/.github/workflows/cmake.yml#L37
> 
> was just a patch on my part since with your change it still works 🤔

Yes, that's what I'm doing. I'm setting the `CMAKE_INSTALL_PREFIX` to a custom location. A location in which we install the thirdparty packages we use. I should've probably mentioned that earlier ...

prince-chrismc:
Okay so it took me a while but I finally got something working. https://github.com/prince-chrismc/jwt-cpp/actions/runs/857066108
I am 100% confused since my understanding of docs did not match the behavoir 🙃 

More test + fixed path https://github.com/Sweco-NL/jwt-cpp/pull/2

prince-chrismc:
Thanks for raising this issue and helping improve our project =)

sweco-nlahoo:
You’re welcome. Thank you for considering my change =)

Van: Chris Mc ***@***.***>
Verzonden: woensdag 19 mei 2021 17:51
Aan: Thalhammer/jwt-cpp ***@***.***>
CC: Hoogesteger, Arjen ***@***.***>; Author ***@***.***>
Onderwerp: Re: [Thalhammer/jwt-cpp] Changed the JWT_CMAKE_FILES_INSTALL_DIR such that it is found on both… (#142)


Thanks for raising this issue and helping improve our project =)

PR #143, opened 2021-05-19
Remove Travis CI

Thalhammer:
Given we do all CI on Github Actions as well, the travis integration adds no addition value.


Thalhammer:
I just noticed that the github actions don't run if you modify non code. This kind of makes sense, however it prevents using them as status checks for merging. As every non code change would be unmergable.

prince-chrismc (approved):

PR #146, opened 2021-06-29
Add JWKS parsing 

Tchinmai7:
builds on top of #112 

closes #112


Thalhammer:
Looks, good, but I have a couple of ideas/additions:
- Liniting fails, (there is a clang-format file in the root that should work with most ide's)
- None of the functions have error_code overloads. We moved the entire library to allow usage without (nonfatal) exception and I'd like to continue this.
- I really dislike the way asymetric keys are passed to the library right now (4 strings for public,private key and a password for each one). I'd like to move to a class abstracting this into a single argument (along with convenience methods) eventually and imho this sounds like a good chance to do so.
- Testing could be more extensive (take a look at the coverage report), but we can add this later once its feature stable.

Tchinmai7:
Thanks for the feedback
1. I'll fix the linting issues right away 
2. For the error_code overloads, I'm thinking of targetting the following functions - what are your thoughts?
```
jwks::get_jwk
jwk::get_algorithm
jwk::get_key_id
jwk::get_x5c
jwk::get_jwk_claim
```
I will add more unit tests and think of a way to abstract away the asymmetric-keys like you suggested

Thalhammer:
I was actually thinking more about the parsing and not the accessors. I am not a computer rn, but you could take a look at the existing code (in particular the verifier) to check which methods have error_codes.

Tchinmai7:
The `parse_jwk` and `parse_jwks` methods invoke the constructors of their respective classes. These methods throw an exception only when the json is malformed, or when there are no keys available. I'm assuming that this is similar to the `decode` method, which also throws an exception in case of invalid input. 

However, if you feel this shouldn't throw an exception and instead use the `error_code` mechanism - i will make the change, just wanted to confirm if that's the intended behavior.

Thalhammer:
> I'm assuming that this is similar to the `decode` method, which also throws an exception in case of invalid input.

I stand corrected on my own library. You are right they do throw on invalid input.
That said it might be a wise idea to move forward to using error_codes there as well.
@prince-chrismc what do you think about this ?

We normally have both overloads, one that takes an error_code and only throws if something really bad happens and one that basically calls the error_code overload and throws if the error code is set.

But I think it might be a good idea to get it merged in some form at least and then iron out this stuff. Shouldn't be a problem as long as it is not part of a release.

Tchinmai7:
I agree - I can send follow-up PR's to change the `parse_jwks`, `parse_jwj`, and `decode` to take in an `error_code`.  I've added a few more unit tests - the classes `jwks` and `jwk` are more or less completely covered now.

Thalhammer:
Ok so I finally came around to actually download the branch and take a look at the jwk specification.
I noticed you only expose a selection of the valid properties (alg, kid and x5c), while
- kty
- use
- key_ops
- x5u
- x5t
- x5t#S256
dont have an accessor. Was this a deliberate decision (e.g. to keep the interface simpler) or an oversight ?
Depending on the kty there can also be other fields like x & y in elliptic curve keys (I don't think they should have direct accessors directly inside the jwk, rather the algorithm should probably get a constructor or static function to load a key from jwk).

It might also be a good idea to check that members required by the spec (e.g. kty) are present and have the correct format.

I don't think it makes sense to add parsing of the keys right now, instead I think we should merge this with the basic structure (which should already be helpfull to people) and follow up with a number of other pr's.
In particular I'd like to do the following:
- Get the base stuff error_code ready and finish transition to error_code interfaces in normal jwt parsing
- Add some sort of "key" type as a member of the algorithm (since they differ vastly) which has functions for parsing various formats (where they make sense), as well as parsing from jwk.
- Add some sort of keychain class which is a list of keys grouped by algorithm (and keyid?)
- Some way to add all keys in a jwk to the allowed algorithms in a verifier.

I am still wondering whether or not to take a look at encryption. Since we already depend on openssl/libressl we would not introduce new dependencies by adding it, but I wonder how many people actually use it (i.e. if its worth the effort).

Tchinmai7:
It was an oversight - I had mistakenly assumed the OP had completely implemented the spec, my bad - I should have verified my assumptions. 

I agree with your list of TODO items - and I think the next step after merging this is to transition to error_code interfaces and stabilize the API before adding more features

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
...
 
+	template<typename json_traits>
+	class jwk {
+		using basic_claim_t = basic_claim<json_traits>;
+		details::map_of_claims<json_traits> jwk_claims;
+
+	public:
+		JWT_CLAIM_EXPLICIT jwk(const typename json_traits::string_type& str) {
+			jwk_claims = details::map_of_claims<json_traits>::parse_claims(str);
+		}
```
prince-chrismc, line 2886: Oh code, it's been a few months I have not worked in C++ but I think we should use initializers

```suggestion
		JWT_CLAIM_EXPLICIT jwk(const typename json_traits::string_type& str) : 
					jwk_claims{details::map_of_claims<json_traits>::parse_claims(str)}
		{}
```

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
...
+
+		/**
+		 * Get jwks claim
+		 * \return Requested claim
+		 * \throw std::runtime_error If claim was not present
+		 */
+		basic_claim_t get_jwk_claim(const typename json_traits::string_type& name) const {
+			return jwk_claims.get_claim(name);
+		}
```
prince-chrismc, line 3048: > dont have an accessor. Was this a deliberate decision (e.g. to keep the interface simpler) or an oversight ?

This was intentional AFAIK... There's not a finite list and it changed depending on the `"alg"` that's used and I am sure the best practices have changed since I looked at this

I personal like that the helpers are public because someone can get any key and even custom ones without it being a nightmare!

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
...
+
 	/**
-	 * Create a verifier using the given clock
-	 * \param c Clock instance to use
-	 * \return verifier instance
-	 */
+   * Create a verifier using the given clock
+   * \param c Clock instance to use
+   * \return verifier instance
+   */
```
prince-chrismc, line 3114: Looks out of place 🤔

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
...
+		details::map_of_claims<json_traits> jwk_claims;
+
+	public:
+		JWT_CLAIM_EXPLICIT jwk(const typename json_traits::string_type& str) {
+			jwk_claims = details::map_of_claims<json_traits>::parse_claims(str);
+		}
+
+		JWT_CLAIM_EXPLICIT jwk(const typename json_traits::value_type& json) {
+			jwk_claims = {json_traits::as_object(json)};
+		}
```
prince-chrismc, line 2890:
```suggestion
		JWT_CLAIM_EXPLICIT jwk(const typename json_traits::value_type& json) :
			jwk_claims{json_traits::as_object(json)}		     {}
```

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
 		}
 	};
 
+	template<typename json_traits>
+	class jwk {
+		using basic_claim_t = basic_claim<json_traits>;
+		details::map_of_claims<json_traits> jwk_claims;
```
prince-chrismc, line 2881: there are no setters currently

```suggestion
		const details::map_of_claims<json_traits> jwk_claims;
```

prince-chrismc (commented):
> You are right they do throw on invalid input.
> That said it might be a wise idea to move forward to using error_codes there as well.
> @prince-chrismc what do you think about this ?

No objections! 👍

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
 		}
 	};
 
+	template<typename json_traits>
+	class jwk {
```
prince-chrismc, line 2879:
```suggestion
	/**
	 * \brief JSON Web Key
	 *
	 * https://tools.ietf.org/html/rfc7517
	 *
	 * A JSON object that represents a cryptographic key.  The members of
         * the object represent properties of the key, including its value.
	 */
	template<typename json_traits>
	class jwk {
```

include/jwt-cpp/jwt.h
```diff
@@ -2873,11 +2875,243 @@ namespace jwt {
...
+		 */
+		basic_claim_t get_jwk_claim(const typename json_traits::string_type& name) const {
+			return jwk_claims.get_claim(name);
+		}
+
+		bool empty() const noexcept { return jwk_claims.empty(); }
+	};
+
+	template<typename json_traits>
+	class jwks {
```
prince-chrismc, line 3054:
```suggestion
	/**
	 * \brief JWK Set
	 *
	 * https://tools.ietf.org/html/rfc7517
	 *
	 * A JSON object that represents a set of JWKs.  The JSON object MUST
         * have a "keys" member, which is an array of JWKs.
         *
         * This container takes a JWKs and simplifies it to a vector of JWKs 
	 */
	template<typename json_traits>
	class jwks {
```

include/jwt-cpp/jwt.h
prince-chrismc: I see https://github.com/Thalhammer/jwt-cpp/pull/146/commits/57b549354a54bfa7e94e96d8bb988ce0c3bf5500 which is good and I think it's even more complete 👍

Thalhammer: Actually I was not talking about the extra fields for each algorithm. I was talking about those I added later. They are in the  jwk standard and can be on any jwk regardless of the algorithm, though pretty much all of them are optional.

Thalhammer:
@Tchinmai7 Thanks for the work you put into this :)

Tchinmai7:
You're welcome, and Thank you very much for this awesome library!

PR #148, opened 2021-07-01
Rework verifier checks

Thalhammer:
Reworked the verifier core to allow more flexibility when checking.
This also removes the special cases for nbf, iat, exp and aud.

Basically I replaced the exact matching against a single value with a more flexible callback for each claim that gets passed some context and can return error codes in case they don't verify correctly. This allows for way more flexible checks like doing deep checks of objects or case insensitive checks.

This is not a final proposal, but more like a proof of the idea. I am absolutely open to improvements/changes even if they completely restructure this.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -20,6 +20,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <functional>
 #include <iterator>
 #include <memory>
 #include <set>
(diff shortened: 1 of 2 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -2606,12 +2607,166 @@ namespace jwt {
 #endif
 	};
 
+	namespace verify_ops {
+		template<typename json_traits>
+		struct verify_context {
+			verify_context(date ctime, const decoded_jwt<json_traits>& j, size_t l)
+				: current_time(ctime), jwt(j), default_leeway(l) {}
```
prince-chrismc, line 2614: This is the only one with a ctor but it shouldn't be required

```suggestion
```

include/jwt-cpp/jwt.h
```diff
@@ -2606,12 +2607,166 @@ namespace jwt {
...
+				}
+				bool matches = false;
+				switch (expected.get_type()) {
+				case json::type::boolean: matches = expected.as_bool() == jc.as_bool(); break;
+				case json::type::integer: matches = expected.as_int() == jc.as_int(); break;
+				case json::type::number: matches = expected.as_number() == jc.as_number(); break;
+				case json::type::string: matches = expected.as_string() == jc.as_string(); break;
+				case json::type::array:
+				case json::type::object:
+					matches = json_traits::serialize(expected.to_json()) == json_traits::serialize(jc.to_json());
```
prince-chrismc, line 2650: 💭 I wonder if we can let the underlying JSON library handle this.. might be a fun test

include/jwt-cpp/jwt.h
```diff
@@ -2606,12 +2607,166 @@ namespace jwt {
...
+					return;
+				}
+				auto c = ctx.jwt.get_payload_claim(ctx.claim_key);
+				if (c.get_type() == json::type::string) {
+					if (expected.size() > 1 || !(expected.empty() || *expected.begin() == c.as_string())) {
+						ec = error::token_verification_error::audience_missmatch;
+						return;
+					}
+				} else {
+					auto jc = c.as_set();
```
prince-chrismc, line 2711: It would be nice to have a template overload for sorted or not...

include/jwt-cpp/jwt.h
```diff
@@ -2606,12 +2607,166 @@ namespace jwt {
...
+		struct is_subset_claim {
+			typename basic_claim<json_traits>::set_t expected;
+			void operator()(const verify_context<json_traits>& ctx, std::error_code& ec) {
+				if (!ctx.jwt.has_payload_claim(ctx.claim_key)) {
+					ec = error::token_verification_error::audience_missmatch;
+					return;
+				}
+				auto c = ctx.jwt.get_payload_claim(ctx.claim_key);
+				if (c.get_type() == json::type::string) {
+					if (expected.size() > 1 || !(expected.empty() || *expected.begin() == c.as_string())) {
```
prince-chrismc, line 2706: This is a complicated expression 🤔 

I think if we don't overload this for matching strings and array its much easier and intuative

include/jwt-cpp/jwt.h
```diff
@@ -2606,12 +2607,166 @@ namespace jwt {
...
+			typename json_traits::string_type claim_key{};
+		};
+
+		/**
+		 * This is the default operation and does case sensitive matching
+		 */
+		template<typename json_traits>
+		struct equals_claim {
+			basic_claim<json_traits> expected;
+			void operator()(const verify_context<json_traits>& ctx, std::error_code& ec) {
```
prince-chrismc, line 2632: I think we can add a lot of `const` in this section

```suggestion
			const basic_claim<json_traits> expected;
			void operator() const (const verify_context<json_traits>& ctx, std::error_code& ec) {
```

include/jwt-cpp/jwt.h
```diff
@@ -2676,8 +2853,34 @@ namespace jwt {
...
+					ec = error::token_verification_error::missing_claim;
+					return;
+				}
+				auto c = ctx.jwt.get_header_claim(ctx.claim_key);
+				if (c.get_type() != json::type::string) {
+					ec = error::token_verification_error::claim_type_missmatch;
+					return;
+				}
+				auto val = c.as_string();
+				std::transform(val.begin(), val.end(), val.begin(), [](char c) { return std::tolower(c); });
```
prince-chrismc, line 2879: didn't know this was required https://datatracker.ietf.org/doc/html/rfc7519#section-5.1

prince-chrismc (commented):
I was worried about a massive API break but I am really pleasantly surprised how nicely this fits in

Overall I think this is excellent, the added flexibility from a function to verify with is extremely helpful, I have a lot of application specific workflow private claims and I can see this being particularly powerful

👍

include/jwt-cpp/jwt.h
```diff
@@ -2712,7 +2917,9 @@ namespace jwt {
 		 * \return *this to allow chaining
 		 */
 		verifier& with_audience(const typename json_traits::string_type& aud) {
-			return with_claim("aud", basic_claim_t(aud));
+			typename basic_claim_t::set_t s;
+			s.insert(aud);
+			return with_audience(s);
```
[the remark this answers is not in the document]
Thalhammer: I don't think there should be a distinction between something like
`"aud":"test"` and `"aud":["test"]`. I might be wrong, but my interpretation of https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3 is that the claim should be accepted as long as the value provided to the verifier is part of the audiences. Thats also how we handled it until now (just in a special case). Your change would mean that `"aud":"test"` matches, but `"aud":["test"]` does not, which imho is wrong.

include/jwt-cpp/jwt.h
Thalhammer: I technically agree and I did not have the constructor in the first place, however for some reason macos refused to compile without it because it thought that the reference could be unassigned (and it is technically right). Not sure if theres a better way to fix this, but this was the simplest way for a POC.

include/jwt-cpp/jwt.h
Thalhammer: Might be a good idea, since most probably have some sort of way to compare json values. However it would be an api breaking change, since we'd need to extend the json traits (and everyone using custom json libraries would need to do the same).

include/jwt-cpp/jwt.h
Thalhammer: The issue is that as far as I understand the rfc its required to accept both array and string values for the audience. We could clean up that if a bit though, to make it easier to read.

include/jwt-cpp/jwt.h
Thalhammer: Neither did I, but reading the standard it appearently is.
It hasn't been an issue before though since we did not provide a way to explicitly check the type (since in 99.9% it won't matter), but somehow people seem to like doing this (theres now a check on the jwt.io library list for specifically this) so I dropped it in along the other changes.

include/jwt-cpp/jwt.h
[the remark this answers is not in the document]
prince-chrismc: I think the issue is

> The interpretation of audience values is generally application specific.

Most APIs I've worked with treat the `"aud"` as a string where an exact match was expected, or it it was a string, it was a FQDN wildcard with regex matching 🤔 neither case is covered here...

Regardless I think keeping the original behavior here is important and I think keeping with it is perfectly correct and inline wiht the spec.

include/jwt-cpp/jwt.h
prince-chrismc: Ahhh it's because without the ctor clang wants you to pass by the init list syntax. This is good 👍

prince-chrismc:
The only comment left for me is https://github.com/Thalhammer/jwt-cpp/pull/148#discussion_r663287108, I'd like to see that added in, otherwise...

LGTM 🚀

include/jwt-cpp/jwt.h
```diff
@@ -2606,12 +2607,166 @@ namespace jwt {
...
+							ec = error::token_verification_error::audience_missmatch;
+							return;
+						}
+					}
+				}
+			}
+		};
+
+		/**
+		 * Checks if the claim is a string and does an case insensitive comparison.
```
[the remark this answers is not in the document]
Thalhammer: Doing this correctly would involve utf-8 handling or am I wrong ?
codecvt should be able to do this in an platform independed way, however its deprecated since C++17 (but will remain in the standard until a replacement is found).
However even with codecvt it means having to deal with locales and weird edge cases, which doesn't sound fun to me.

include/jwt-cpp/jwt.h
Thalhammer: Thinking about it we might be able to detect if the "equals" method is part of the json_traits with some template magic and switch between the implementations we have now and the traits method using specialization or enable_if, without breaking everyones code. But we should probably change that everywhere we compare json values then.

include/jwt-cpp/jwt.h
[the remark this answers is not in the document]
prince-chrismc: 💯 I agree, I think just adding a small disclaimer is more then enough

prince-chrismc (approved):

Issue #150, opened 2021-07-22
Eliminate deprecations warnings with OpenSSL 3


NFM-8:
Currently, build against OpenSSL 3.0.0 release candidate fails because many deprecated OpenSSL functions is used.

For more information, see https://www.openssl.org/docs/manmaster/man7/migration_guide.html

Thalhammer:
Ahhhh what happened to OpenSSL 2.x ?
Anyway thanks for bringing it up, I hadn't even realized that OpenSSL 3 was a thing.
We will ofc support it. That said it might take a while to set up a dev environment that supports such recent versions.

prince-chrismc:
Tests passing 👍 https://github.com/Thalhammer/jwt-cpp/actions/runs/1127485043

In time we'll fix the deprecation warnings

Thalhammer:
I don't think we should close this. For many people (including me) a deprecation warning actually breaks the build, since they build with Werror on. Sure those could ignore the warning for jwt-cpp, but I don't feel like that's a valid recommendation. I'd rather only close this if we either ignore the warnings in our code or (much better) fix them.

prince-chrismc:
That's reasonable! I read "support" and that was added, but yes we can go the extra mile

PR #153, opened 2021-08-08
Fix tests for OpenSSL 3.0.0

kleinmrk:
OpenSSL 3.0.0 changes the signature of
`int PEM_write_bio_PUBKEY(BIO *bp, EVP_PKEY *x);` to
`int PEM_write_bio_PUBKEY(BIO *bp, const EVP_PKEY *x);`,

and 
`int PEM_write_bio_X509(BIO *bp, X509 *x);` to
`int PEM_write_bio_X509(BIO *bp, const X509 *x);`.

The following tests failed since simulation of OpenSSL errors did not take place:
- `OpenSSLErrorTest.ExtractPubkeyFromCert`
- `OpenSSLErrorTest.ConvertCertBase64DerToPem`
- `OpenSSLErrorTest.ConvertCertBase64DerToPemErrorCode`
- `OpenSSLErrorTest.ECDSACertificate`
- `OpenSSLErrorTest.EdDSACertificate`

### DIFF
--- a/.github/actions/install/openssl/action.yml
+++ b/.github/actions/install/openssl/action.yml
@@ -1,0 +1,19 @@
+name: Install OpenSSL
+description: Install and setup OpenSSL for linking and building test application
+inputs:
+  version:
+    description: The desired OpenSSL version to install
+    required: false
+    default: "3.0.0-beta2"
+runs:
+  using: composite
+  steps:
+    - run: |
+        cd /tmp
+        wget http://artfiles.org/openssl.org/source/openssl-${{ inputs.version }}.tar.gz
+        tar -zvxf /tmp/openssl-${{ inputs.version }}.tar.gz
+        cd openssl-${{ inputs.version }}
+        ./config --prefix=/tmp --libdir=lib
+        make -j $(nproc)
+        sudo make -j $(nproc) install_sw
+      shell: bash
--- a/.github/actions/install/openssl/action.yml
+++ b/.github/actions/install/openssl/action.yml
@@ -4,14 +4,14 @@ inputs:
   version:
     description: The desired OpenSSL version to install
     required: false
-    default: "3.0.0-beta2"
+    default: "openssl-3.0.0-beta2"
 runs:
   using: composite
   steps:
     - run: |
         cd /tmp
-        wget http://artfiles.org/openssl.org/source/openssl-${{ inputs.version }}.tar.gz
-        tar -zvxf /tmp/openssl-${{ inputs.version }}.tar.gz
+        wget https://github.com/openssl/openssl/archive/refs/tags/${{ inputs.version }}.tar.gz
+        tar -zxf /tmp/${{ inputs.version }}.tar.gz
         cd openssl-${{ inputs.version }}
         ./config --prefix=/tmp --libdir=lib
         make -j $(nproc)
--- a/.github/workflows/openssl.yml
+++ b/.github/workflows/openssl.yml
@@ -10,7 +10,7 @@ on:
       - "include/jwt-cpp/**"
       - "tests/cmake/**"
       - ".github/actions/**"
-      - ".github/workflows/cmake.yml"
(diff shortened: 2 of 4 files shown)


prince-chrismc:
Is this related to #150? Could you comment how bad the usage of deprecated methods is?

kleinmrk:
I see this as a first step towards OpenSSL 3.0.0 support. I just wanted to give it a shot with the release candidate and figured that the tests are failing. Now that the tests are passing one can start getting rid of the deprecated API calls by using higher level API.

Regarding how bad deprecated methods are:
From https://www.openssl.org/docs/manmaster/man7/migration_guide.html:
> Use of the low level APIs has been informally discouraged by the OpenSSL development team for a long time. However in OpenSSL 3.0 this is made more formal. All such low level APIs have been deprecated. You may still use them in your applications, but you may start to see deprecation warnings during compilation (dependent on compiler support for this). Deprecated APIs may be removed from future versions of OpenSSL so you are strongly encouraged to update your code to use the high level APIs instead.

I think that at the moment using this low level API does not pose a big threat, but it is just matter of time till they start dropping deprecated functions.

Thalhammer:
Looks good to me. Regarding the deprecation: While I agree that we should want to fix them (e.g. I compile with Werror, so a deprecation warning is effectively a broken build), I don't think there's a huge hurry. OpenSSL said they want to keep it as a drop in replacement for existing programs which means they can't really remove them until all major software is fixed. Given the widespread use of openssl, I am pretty sure it will take a loooong time until there's an actual threat of them being removed.

prince-chrismc:
> The following tests failed since simulation of OpenSSL errors did not take place:

We are missing unit tests in out GitHub Actions

kleinmrk:
@prince-chrismc I added another commit that introduces OpenSSL CI workflow similar to LibreSSL CI. It builds, links and runs unit tests with different versions of OpenSSL.

.github/workflows/openssl.yml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+    paths:
+      - "CMakeLists.txt"
+      - "include/jwt-cpp/**"
+      - "tests/cmake/**"
+      - ".github/actions/**"
+      - ".github/workflows/cmake.yml"
```
prince-chrismc, line 13:
```suggestion
      - ".github/workflows/openssl.yml"
```

prince-chrismc:
Thank you so much! I had the intention of working on this however life had other plans.

❤️ Greatly appreciate the contribution 

Emergency vet visit for dog 🤕 (the sock is so he couldn't like the cut)

prince-chrismc:
https://github.com/Thalhammer/jwt-cpp/pull/153/checks?check_run_id=3307203891#step:8:267

All the tests passed + plus it picked up the correct version it looks like we are in business!

.github/actions/install/openssl/action.yml
```diff
@@ -0,0 +1,19 @@
...
+  version:
+    description: The desired OpenSSL version to install
+    required: false
+    default: "3.0.0-beta2"
+runs:
+  using: composite
+  steps:
+    - run: |
+        cd /tmp
+        wget http://artfiles.org/openssl.org/source/openssl-${{ inputs.version }}.tar.gz
```
prince-chrismc, line 13: I have no issue with using a mirror however perhaps using GitHub is better for actions? 

```suggestion
        wget https://github.com/openssl/openssl/archive/refs/tags/openssl-${{ inputs.version }}.tar.gz
```

kleinmrk: Makes sense. I updated links, however,  the versions are now referenced as `openssl-3.0.0-beta2` and `OpenSSL_1_1_1k`. I don't know whether there is an easy support for something like maps in action files.

prince-chrismc (approved):

PR #156, opened 2021-08-16
🔐  Adding support for `"alg"` ES256K

### LINKED ISSUE #79: Add support for latest algorithms
There's two new algorithms that have been adopted to the JWT universe, specifically JOSE
- ~[EdDSA](https://tools.ietf.org/html/rfc8037)~
- [ES256K](https://tools.ietf.org/html/draft-ietf-cose-webauthn-algorithms-05#section-3)

jwt-cpp could be expanded to support these as well

prince-chrismc:
resolves #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,12 @@ jwt-cpp supports all the algorithms defined by the specifications. The modular d
 
 For completeness, here is a list of all supported algorithms:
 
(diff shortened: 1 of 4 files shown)


Thalhammer (approved):
Looks good to me, but we should probably verify a given key actually has the right curve for the selected algorithm. But we can do that in a later commit since I consider this user error and I think we should rework the key handling anyway.

PR #157, opened 2021-08-17
📦 List available package managers

prince-chrismc:
and fix the ugly looking badge for openssl =D 

0.5.1 ➡️ https://github.com/microsoft/vcpkg/pull/19618 https://github.com/conan-io/conan-center-index/pull/6893

### DIFF
--- a/.github/workflows/openssl.yml
+++ b/.github/workflows/openssl.yml
@@ -5,33 +5,28 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
-    paths:
-      - "CMakeLists.txt"
-      - "include/jwt-cpp/**"
-      - "tests/cmake/**"
-      - ".github/actions/**"
-      - ".github/workflows/openssl.yml"
 
 jobs:
   with-openssl:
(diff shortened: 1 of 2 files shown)


Thalhammer:
We also need to update the jwt.io website with the current support table.
Hunter support might be worth looking into. It is a package manager written entirely in cmake and works really well. I've used it in a couple of internal projects lately. Support should be pretty easy to integrate.

prince-chrismc:
How do we update the jwt.io website?

Thalhammer:
> How do we update the jwt.io website?

If I remember correctly we need to fork the website repo, update a json file and create a pull request, which is normally accepted without further questions.
Imma go ahead and do so, since I already added it back then.

Thalhammer (approved):
Do we want to support hunter with this commit, or with a different pr or not at all ?
Cause if we intend to, we would need to add some files and change the cmakelists a bit.

prince-chrismc:
Separate PR since there's more modifications

PR #163, opened 2021-09-01
Integrate Hunter package manager support

Thalhammer:
Initial support for using jwt-cpp safely inside hunter based projects. If the project is included inside a hunter project it will use hunter to resolve its dependencies (at the moment only openssl and gtest) and warn if it can't.

If it would not do so it might result in jwt-cpp code linking against a system openssl lib, while other code uses a (probably different) hunter one.

In addition this will allow inclusion of jwt-cpp as a hunter package in the future.


prince-chrismc (approved):
Integration is really clean

PR #165, opened 2021-09-07
Add support for Boost.JSON traits

### LINKED ISSUE #164: static_assert causes Boost.JSON integration to fail
I have written an example on how to integrate Boost.JSON with this library. However, the following static assert:

```cpp
static_assert(
    details::is_valid_json_types<typename json_traits::value_type,
                                 typename json_traits::string_type,
                                 typename json_traits::object_type,
                                 typename json_traits::array_type>::value,
    "must staisfy json container requirements");
```
in `jwt.h` line 1872 causes every `set_*` function to error when creating a JWT.

The produced error is:

```
In template: static_assert failed due to requirement 'is_valid_json_object<boost::json::value, std::basic_string<char, std::char_traits<char>, std::allocator<char>>, boost::json::object>::value' "object_type must be a string_type to value_type container"
```
If I remove that static assert then I am able to successfully create a token. I have also been able to verify the resulting token using a separate library.

Before submitting a pull request to add this example to the code base, I would appreciate some discussion/insight as to why this is occurring.

Here is the complete example of the Boost.JSON integration.

```cpp
#define JWT_DISABLE_PICOJSON

#include <iostream>
#include <boost/json.hpp>

#include "jwt.h"

namespace json = boost::json;           // from <boost/json/src.hpp>

struct boost_json_traits {
    // Type Specifications
    using value_type = json::value;      // The generic "value type" implementation, most libraries have one
    using object_type = json::object;    // The "map type" string to value
    using array_type = json::array;      // The "list type" array of values
    using string_type = std::string;      // The "list of chars", must be a narrow char
    using number_type = double;       // The "precision type"
    using integer_type = int64_t;       // The "integral type"
    using boolean_type = bool;         // The "boolean type"

    // Translation between the implementation notion of type, to the jwt::json::type equivalent
    static jwt::json::type get_type(const value_type &val) {
        using jwt::json::type;

        if (val.is_object())
            return type::object;
        if (val.is_array())
            return type::array;
        if (val.is_string())
            return type::string;
        if (val.is_double())
            return type::number;
        if (val.is_int64() || val.is_uint64())
            return type::integer;
        if (val.is_bool())
            return type::boolean;

        throw std::logic_error("invalid type");
    }

    // Conversion from generic value to specific type
    static object_type as_object(const value_type &val) {
        if (val.is_object()) return val.as_object();
        throw std::bad_cast();
    }
    static array_type as_array(const value_type &val) {
        if (!val.is_array()) return val.as_array();
        throw std::bad_cast();
    }
    static string_type as_string(const value_type &val) {
        if (!val.is_string()) return val.as_string().c_str();
        throw std::bad_cast();
    }
    static number_type as_number(const value_type &val) {
        if (!val.is_double()) return val.as_double();
        throw std::bad_cast();
    }
    static integer_type as_int(const value_type &val) {
        if (val.is_int64()) return val.as_int64();
        if (val.is_uint64()) return val.as_uint64();
        throw std::bad_cast();
    }
    static boolean_type as_bool(const value_type &val) {
        if (!val.is_bool()) return val.as_bool();
        throw std::bad_cast();
    }

    // serialization and parsing
    static bool parse(value_type &val, const string_type& str) {
        val = json::parse(str);
        return true;
    }
    static string_type serialize(const value_type &val) {
        return json::serialize(val);
    }
};

int main() {
    std::string pub_key = "YOUR PUB KEY HERE";
    std::string priv_key = "YOUR PRIV KEY HERE";

    auto token = jwt::create<boost_json_traits>()
            .set_issuer("iss")
            .set_type("JWT")
            .set_issued_at(std::chrono::system_clock::now())
            .set_expires_at(std::chrono::system_clock::now() + std::chrono::seconds{3600})
            .set_payload_claim("token_use", "access")
            .sign(jwt::algorithm::rs256(pub_key, priv_key, "", ""));

    std::cout << "token:\n" << token << std::endl;
}
```

prince-chrismc:
- requires the C++17 "standalone" version of the library
- it can **not** be used from a full boost installation 

resolve #164 

When I played with adding the C++11 full boost version, I found some missing assertions about the `string_type` which you can see were added


include/jwt-cpp/jwt.h
```diff
@@ -1847,7 +1847,8 @@ namespace jwt {
 				is_detected<has_mapped_type, object_type>::value &&
 				std::is_same<typename object_type::mapped_type, value_type>::value &&
 				is_detected<has_key_type, object_type>::value &&
-				std::is_same<typename object_type::key_type, string_type>::value &&
+				(std::is_same<typename object_type::key_type, string_type>::value ||
+				 std::is_constructible<typename object_type::key_type, string_type>::value) &&
```
prince-chrismc, line 1851: This is the fix

tests/boost/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 3.18)
```
prince-chrismc, line 1: https://cmake.org/cmake/help/latest/command/file.html#archive-extract

The upstream CMake scripts do no install the headers of config files so I opted to keep the "install" here instead of making an action for it

PR #167, opened 2021-09-11
⬆️ Bump tested SSL options

prince-chrismc:


prince-chrismc:
@Thalhammer may you please remove the required checks for the older versions of LibreSSL? In a future PR I am working on they'll get renamed might not be worth updating them

Thalhammer:
> @Thalhammer may you please remove the required checks for the older versions of LibreSSL?

I think I will just remove all of them, since its only us two that can merge anyway.

Issue #166, opened 2021-09-09
Clang detects potential memory leak.


ttosbt:
Howdy, I am new to adding issues to public libraries my projects use, so please have patience with me.  :^)

Using clang-tidy (LLVM 10.0.1), during build I receive the error clang-analyzer-cplusplus.NewDeleteLeaks.  Looks to be related to the way Pico is handling the string, but I am not 100% on that.  This is with an RSA 256 token.

The issue is encountered at jwt.h:2934

I am on an x86_64 Linux machine running kernel 4.18.5.  CentOS 7.7.1908.

It should be noted that instead of running the example code directly I split up the creation of the verifier and any follow on validations that need to be added based on configuration.

For example in my header:

`jwt::verifier<jwt::default_clock, jwt::picojson_traits> the_verifier{jwt::verify()};`  (Which admittedly seems ridiculously complicated to define this way...but it's the best I could do given the examples all use auto and everything is constructed in place and seemingly never reused.)

Then, later in my implementation:

```
if (verify_issuer) {
    the_verifier.with_issuer(the_issuer_from_configuration_to_use);
}
```

I don't want to just run the code in the example file because I won't know the issuer that's come in until it's come in and I am dynamically building objects to handle connections which are created at start and put into a pool for use/reuse.  Running the creation every time seemed wasteful and less performant than just creating a verifier and configuring it on the fly with configuration as needed.

Thanks in advance for any help you may be able to provide.

Thalhammer:
> I split up the creation of the verifier and any follow on validations that need to be added based on configuration
This should be fine and is a supported use case.

> ridiculously complicated to define this way
Thats often the price you pay for flexibility. However since you initialize it directly you should be fine using auto as the variable type even in a header.

Assuming you mean line 2934 in the current master https://github.com/Thalhammer/jwt-cpp/blob/ac0424b115721e4066d2fb99f72ba0cd58759882/include/jwt-cpp/jwt.h#L2934 I can't see how it should leak (assuming picojson is fine, which we do not maintain) and running the test suit in valgrind says everything is deleted, so none of the test cases leak anything. Given this I am tempted to say its a false positive, however just to be sure, can you try actually running the relevant code with something like asan or valgrind to check if it does leak ?

ttosbt:
Cool, that was a fast turn around!  Thanks for that.  Appreciate the feedback.

[...]

PR #168, opened 2021-09-11
Adding clang-tidy, cmake-format, and ASAN tests

prince-chrismc:
Playing with clang-tidy, we already had the rule mentioned in #166... running the static analyzer from CMake reproduced the report in many places... I suspect the macro and `new`/`delete` separation clang-tidy is just unable to understand.

- Running the problematic examples + tests with the ASAN enabled reported no problems
- Running examples with nlohmann/json and jsoncons did not have any warnings related to the memory leak


include/jwt-cpp/jwt.h
```diff
@@ -2603,7 +2603,8 @@ namespace jwt {
 		typename json_traits::string_type sign(const Algo& algo, Encode encode, std::error_code& ec) const {
 			// make a copy such that a builder can be re-used
 			typename json_traits::object_type obj_header = header_claims;
-			if (header_claims.count("alg") == 0) obj_header["alg"] = typename json_traits::value_type(algo.name());
+			if (header_claims.count("alg") == 0)
+				obj_header.emplace(std::make_pair("alg", typename json_traits::value_type(algo.name())));
```
prince-chrismc, line 2607:
```suggestion
			if (header_claims.count("alg") == 0) obj_header["alg"] = typename json_traits::value_type(algo.name());
```

PR #172, opened 2021-09-15
Remove outdated debian packaging files

prince-chrismc:
If we decided to keep them, it's better to write tests for them maybe automate the process?


Thalhammer:
AFAIK if we wanted to do deb files we should probably use cpack instead.

prince-chrismc:
Agreed!

PR #184, opened 2021-11-17
Bump WolfSSL to v5.0.0

prince-chrismc:
Move to the latest release since we were previously testing against a commit on master which had new code that's just been released


Thalhammer (approved):
Looks good to me.

PR #185, opened 2021-11-18
simplify reading of EC keys

kleinmrk:
This MR slightly changes how EC keys are read.

Currently, both private and public keys are read (if provided), but the public key has no effect and the internally used openssl structure is overridden by the private part. Thus the public key has no effecton the final ecdsa object other than possibly throwing exceptions in the constructor.

This MR reads the keys the same way as they are read for RSA key, i.e., prefer private key if present, and completely ignore public part.

I also switched from `PEM_read_bio_EC_PUBKEY` to `PEM_read_bio_PUBKEY`, and from `PEM_read_bio_ECPrivateKey` to `PEM_read_bio_PrivateKey`. This change made it obvious that one could use the same code for reading RSA and EC keys. The functions only differ in type of errors they report, e.g., `error::ecdsa_error::create_mem_bio_failed` vs. `error::rsa_error::create_mem_bio_failed`.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -639,8 +639,8 @@ namespace jwt {
 		 * \param pw		Password used to decrypt certificate (leave empty if not encrypted)
 		 * \param ec		error_code for error_detection (gets cleared if no error occures)
 		 */
-		inline std::shared_ptr<EVP_PKEY> load_public_ec_key_from_string(const std::string& key,
-																	 const std::string& password, std::error_code& ec) {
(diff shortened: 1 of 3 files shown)


prince-chrismc:
We should deprecate/remove the older helpers and rename them to have RSA in the name.

https://github.com/Thalhammer/jwt-cpp/blob/master/include/jwt-cpp/jwt.h#L581
https://github.com/Thalhammer/jwt-cpp/blob/b8e55e256029c912ad3b30ba602929e236ffa941/include/jwt-cpp/jwt.h#L625

prince-chrismc (approved):

PR #199, opened 2022-01-23
Make it possible to build with OpenSSL 1.0.1

alexeyserbin:
This patch updates the code in the jwt-cpp/jwt.h header and relaxes the
requirement for OpenSSL's version in the top-level CMakeLists.txt to
allow for compiling with OpenSSL 1.0.1.  The patch works around the
non-constantness of the 'sig' parameter in the signature of the
EVP_DigestVerifyFinal() function in versions of OpenSSL prior to 1.0.2.

I verfied it's now possible to compile with OpenSSL 1.0.1, but I guess
it might be possible to compile even with OpenSSL 1.0.0 after tweaking
the version requirement in CMakeLists.txt.

The new EVP_DigestVerifyFinal() signature was introduced in [1] and then
cherry-picked into the main branch as [2].  The update [1] doesn't
contain any functional change per se: even prior to the update,
EVP_DigestVerifyFinal() implementation was de facto treating the 'sig'
parameter as if it were immutable [3].

[1] https://github.com/openssl/openssl/commit/1abfa78a8ba714f7e47bd674db53dbe303cd1ce7
[2] https://github.com/openssl/openssl/commit/0f7fa1b1904d0f5e997f0b150321ef6d3492651b
[3] https://github.com/openssl/openssl/blob/27007233db5d6f8b91ed474c4e09dd7014871cc6/crypto/evp/m_sigver.c#L163-L187


prince-chrismc:
Seems like such  a small change 🤔 I was going to say 1.0.1 has been end of life for 5 years but its tempting to add support.

I am not sure how many other would benefit given we are at OpenSSL 3.x  now.

Thanks a lot for sharing it's really interesting

Thalhammer (changes requested):
While I agree that 1.x is end of life, its still widely used and certainly will be for a long time. Heck some systems are still at 0.9.x. Its a tiny change and maybe someone somewhere happens to run that exact version and gets a happy smile.

Linting fails, but apart from that feel free to merge it.

prince-chrismc:
https://github.com/Thalhammer/jwt-cpp/runs/4933166002?check_suite_focus=true#step:8:205 🚀

PR #171, opened 2021-09-14
Add Wolfssl support

prince-chrismc:

PR #202, opened 2022-01-26
Update documentation layout

prince-chrismc:
Preview: https://prince-chrismc.github.io/jwt-cpp/

I noticed some reference with traits were not included and once I got started I saw there are improved "modern" layout so I'd love any thoughts if this is a good direction 🙏 

I also find the landing README a bit long so I moved two sections into a separate page. Plus added the things that came out from the "using wolfssl" cconverstation from #171


Doxyfile
```diff
@@ -32,19 +32,19 @@ DOXYFILE_ENCODING      = UTF-8
...
 
-PROJECT_NAME           = "JWT-C++"
+PROJECT_NAME           = JWT-CPP
 
 # The PROJECT_NUMBER tag can be used to enter a project or revision number. This
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 0.5.0
+PROJECT_NUMBER         = v0.5.0
```
Thalhammer, line 41: I think we are already at v0.5.1 with 0.6 in prerelease, so we need to update this.

docs/faqs.md
```diff
@@ -0,0 +1,34 @@
+# Frequently Asked Questions
+
+## The generated JWT token can be decode, is this correct and secure?
```
Thalhammer, line 3: decode is missing a d at the end => "can be decoded". I'd also refrase the paragraph a bit. First of all the hash part should probably be encryption since a hash can not be reversed, its a oneway operation. Might also be a good idea to fuse this and the next paragraph. I'd write something along the lines of:
```
This is the expected behaviour. While the integrity of tokens is ensured by the generated/verified hash,
the contents of the token are only encoded and not encrypted. This means you can be sure the token
has not been modified by an unauthorized party, but you can not store confidential information in it. 
Anyone with access to the token can read all the claims you put into it. They can however not modify
them unless they have the (private or symetric) key used to generate the token. If you need to put
confidential information into it, you have to encrypt it beforehand using some other method, though
even then I recommend against doing it. Instead generate a random id and store the data on your
server, using the id to look it up whenever you need.
```

docs/faqs.md
```diff
@@ -0,0 +1,34 @@
...
+## Can this library encrypt/decrypt claims?
+
+No it does not, see [Issue 115](https://github.com/Thalhammer/jwt-cpp/issues/115) for more details.
+More importantly you probably dont want to be using JWTs for anything sensitive. Read [this](https://stackoverflow.com/a/43497242/8480874)
+for more.
+
+## Why are my tokens immediately expired?
+
+If you are generating tokens that seem to immediately expire, you are likely not using UTC. Specifically,
+if you use `get_time` to get the current time, it likely uses localtime, while this library uses UTC,
```
Thalhammer, line 19: This repeats its self.

Thalhammer (changes requested):
Sweeeet, where can I learn this sorcery?

I added some notes to improve the wording, which might be worth changing while we are at it.

prince-chrismc:
> where can I learn this sorcery?

Google 😆 https://www.google.com/search?q=doxygen+themes and then checking the first page for good options

The one I found have an easy installation https://jothepro.github.io/doxygen-awesome-css/index.html#autotoc_md8

It only takes 3 lines

https://github.com/Thalhammer/jwt-cpp/blob/559520b4070bd676fc6ced09b5109235834b755a/.github/workflows/documentation.yml#L12-L13

and 

https://github.com/Thalhammer/jwt-cpp/blob/559520b4070bd676fc6ced09b5109235834b755a/Doxyfile#L1195

docs/faqs.md
prince-chrismc:
```suggestion
## The generated JWT token can be decoded, is this correct and secure?
```

Doxyfile
prince-chrismc:
```suggestion
PROJECT_NUMBER         = v0.5.1
```

prince-chrismc:
One more styling question

How do you feel about the diagram that are being generated?
Example: https://prince-chrismc.github.io/jwt-cpp/structjwt_1_1algorithm_1_1es256k.html

Thalhammer:
> One more styling question
> 
> How do you feel about the diagram that are being generated? Example: https://prince-chrismc.github.io/jwt-cpp/structjwt_1_1algorithm_1_1es256k.html

Look good to me

Thalhammer (approved):

docs/ssl.md
```diff
@@ -0,0 +1,19 @@
...
+Here's an example:
+
+```sh
+cmake . -DJWT_SSL_LIBRARY:STRING=wolfSSL 
+```
+
+## Notes
+
+JWT-CPP relies on the OpenSSL API, as a result both LibreSSL and wolfSSL need to include their respective compatability layers.
+Most system already have OpenSSL so it's important to make sure when compiling you application only the correct one is used.
```
prince-chrismc, line 19:
```suggestion
Most system already have OpenSSL so it's important to make sure when compiling your application it only includes one. Otherwise you may have missing symbols when linking.
```

docs/ssl.md
```diff
@@ -0,0 +1,19 @@
+# Cryptography Libraries
+
+The underlying cryptography libraries describe [here](../README.md#ssl-compatibility) can be selected when configuring CMake by
+explicity setting `JWT_SSL_LIBRARY` to one of three values
```
prince-chrismc, line 4:
```suggestion
The underlying cryptography libraries describe [here](../README.md#ssl-compatibility) can be selected when 
configuring CMake by explicitly setting `JWT_SSL_LIBRARY` to one of three values. The default is to use OpenSSL.
```

Thalhammer (approved):

PR #221, opened 2022-04-03
Support loading jwk from json

kleinmrk:
This MR enables loading keys from json to `jwk` objects. It also adds an interface to the verifier class which accepts such keys and uses them as appropriate, based on the `kid` and `alg` claims, during token verification.

So far only RSA and oct keys can be loaded from json to `jwk`

- [x]  RSA keys
- [ ] EC keys
- [x] oct keys (HMAC)


kleinmrk:
This is still just a draft, there is much to do regarding error handling (currently I just lazily throw `std::runtime_error` all over the place) but I would be happy to hear your feedback, whether this PR goes the right direction.

include/jwt-cpp/jwt.h
```diff
@@ -3267,219 +3705,34 @@ namespace jwt {
...
-			for (auto& c : claims) {
-				ctx.claim_key = c.first;
-				c.second(ctx, ec);
-				if (ec) return;
+			if (!key_found) {
+				if (algs.count(algo) == 0) {
+					ec = error::token_verification_error::wrong_algorithm;
+					return;
+				}
+				algs.at(algo)->verify(data, sig, ec);
```
Thalhammer, line 3730: I know this is my own code, but since we are already at changing it, this would be better:
```cpp
auto alg = algs.find(algo);
if(alg == algs.end()) {
      ec = error::token_verification_error::wrong_algorithm;
      return;
}
alg->second->verify(data, sig, ec);
```
Prevents the duplicate lookup.

include/jwt-cpp/jwt.h
```diff
@@ -3247,6 +3673,18 @@ namespace jwt {
 			return *this;
 		}
 
+		verifier& allow_key(const jwt::jwk<json_traits>& key) {
+			std::string keyid = "";
+			if (key.has_key_id()) {
+				keyid = key.get_key_id();
+				typename keysets::const_iterator it = keys.find(keyid);
```
Thalhammer, line 3680: `auto` is your friend

Thalhammer (commented):
I added a bunch of small nitpicks and one thing I really think we should change.
One of jwt-cpp's strengths has been that its trivial to add custom algorithms and this has indeed been used in academics to research new ones.
The jwks part doesn't take that into account at all. While I get that the set of algorithms currently allowed in jwks is fixed, it might be useful to provide a way to add custom ones for research or if someone needs support of future algorithms before we add it.

Apart from that I have no major complains. Obviously documentation is non existant yet, but thats fine for a draft, just something we should keep an eye on before finishing it.

EDIT: Macos and no base seems to fail, so we need to take the user provided base64 function into account.

kleinmrk:
@Thalhammer 
> One of jwt-cpp's strengths has been that its trivial to add custom algorithms and this has indeed been used in academics to research new ones.
> The jwks part doesn't take that into account at all. While I get that the set of algorithms currently allowed in jwks is fixed, it might be useful to provide a way to add custom ones for research or if someone needs support of future algorithms before we add it.

I looked into how to make it possible to extend JWKs by custom algorithm and I ended up exposing [`algo_base` and `algo` structures](https://github.com/kleinmrk/jwt-cpp/pull/7/files#diff-be2369ece29c0945fc04ffe13a39767b7841fc2fabd3df209a26b3bb381a2fb2R3417-R3428) so the users can register their own algorithms [like this](https://github.com/kleinmrk/jwt-cpp/pull/7/files#diff-7b94d22894c611d07cf7fb02d7eca101222eb7eeeb17730f37326b005b6cd636R97-R98). I don't know whether this is acceptable but I am out of other ideas. If you don't like the approach then please stop me now :D

include/jwt-cpp/jwt.h
```diff
@@ -3559,11 +3418,464 @@ namespace jwt {
...
+				ec = error::token_verification_error::wrong_algorithm;
+				return nullptr;
+			}
+
+			if (alg_name == "RS256") {
+				return std::unique_ptr<algo<jwt::algorithm::rs256>>(
+					new algo<jwt::algorithm::rs256>(jwt::algorithm::rs256(key.get_pkey())));
+			} else if (alg_name == "RS384") {
+				return std::unique_ptr<algo<jwt::algorithm::rs384>>(
+					new algo<jwt::algorithm::rs384>(jwt::algorithm::rs384(key.get_pkey())));
```
prince-chrismc, line 3614: Please use `make_unique`

include/jwt-cpp/jwt.h
```diff
@@ -3559,11 +3418,464 @@ namespace jwt {
...
+			std::error_code ec;
+			verify(jwt, ec);
+			error::throw_if_error(ec);
+		}
+		/**
+		 * Verify the given token.
+		 * \param jwt Token to check
+		 * \param ec error_code filled with details on error
+		 */
+		void verify(const decoded_jwt<json_traits>& jwt, std::error_code& ec) const {
```
prince-chrismc, line 3845: A lot of code moved so the diff is very confusing. It hard to tell what is new 🙊

Issue #213, opened 2022-02-26
[bug] Unable to use picojson installed to `/usr/include/picojson.h`


olifre:
**Describe the bug**
`picojson` defaults to installation into `<prefix>/include/picojson.h`. 
However, `jwt-cpp` tries to include from `<includepath>/picojson/picojson.h` which will not be used by a standard `picojson` installation. 

**Desktop (please complete the following information):**
 - OS: Gentoo Linux with `picojson` installed from [upstream](https://github.com/kazuho/picojson) via system package `dev-cpp/picojson`-
 - Compiler: `gcc (Gentoo 11.2.0 p1) 11.2.0`

prince-chrismc:
You are absolutely correct, https://github.com/kazuho/picojson/blob/111c9be5188f7350c2eac9ddaedd8cca3d7bf394/Makefile#L21

Do you have any suggestions on how we can support both?

olifre:
I believe the most flexible approach would be (but that needs some more changes):

* Make `JWT_EXTERNAL_PICOJSON` a bit more flexible, i.e. in case it is set, allow the user to specify a `PICOJSON_INCLUDE_PATH` from outside instead of `find_package`. 

* Embed that path (which could default to `picojson/` so it works with the bundled version) into the `include` statements. That could be done via a preprocessor macro and a header config file created by `CMake` for example.

olifre:
> jwt-cpp could provide a custom CMake module that searches both locations and setups the target correctly?

That sounds like a good solution :-). 

My experience with hunter, conan or vcpkg is sadly zero, I have first heard of these in this project. As a quick hack, I have for now used the following approach in Gentoo packaging:
https://github.com/gentoo/guru/blob/47471944d857275e82983460b6cd720257510df7/dev-cpp/jwt-cpp/jwt-cpp-0.6.0.ebuild#L31-L38

I'm not sure I'll have sufficient time for a PR in the near future, but I believe the approach is a good one :+1: .

prince-chrismc:
Note for whoever gets to this first, #195 is closely related and we should write some docs to help the community

PR #222, opened 2022-04-05
Fix Conan Example

csegarragonz:
As mentioned in a related issue, the Conan example is broken. It uses version `0.4.0` that still installs `picojson/1.3.0` as a dependency.

[For code versions greater than `0.5.0`](https://github.com/conan-io/conan-center-index/blob/master/recipes/jwt-cpp/all/conanfile.py#L29), `picojson` is no longer installed as a dependency, and the provided example fails to build.

In this PR I provide a minimal fix, but you may prefer to change the Conan recipe altogether.

### DIFF
--- a/example/conan/conanfile.txt
+++ b/example/conan/conanfile.txt
@@ -1,5 +1,6 @@
 [requires]
-jwt-cpp/0.4.0
+jwt-cpp/0.6.0
+picojson/1.3.0
 
 [generators]
 cmake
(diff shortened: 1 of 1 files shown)


prince-chrismc:
> As mentioned in a related issue,

Sorry for not responding, I am "taking time off" so I was not able to answer https://github.com/Thalhammer/jwt-cpp/issues/213#issuecomment-1087744619

But I am very glad you found the issue! I am super grateful you've gone above and beyond to contribute back and improve our example! 🚀 💟 

> `picojson` is no longer installed as a dependency... but you may prefer to change the Conan recipe altogether.

There a lot of context behind it but this is the intended usage... https://github.com/conan-io/conan-center-index/pull/4644#discussion_r580966605

PR #225, opened 2022-04-12
Make jwt::decoded_jwt moveable

### LINKED ISSUE #224: Make jwt::decoded_jwt moveable
### What would you like to see added?

jwt::decoded_jwt to be moveable

### Additional Context

By having the token member of jwt::decoded_jwt as a const, a object of jwt::decoded_jwt cannot be moved. I think it would be nice if the token could be moved.

Ka0o0:
This removes the const specifier from the token member class so that objects of type jwt::decoded_token can be moved. IMO it is safe to do so as the variable is only exposed via `get_token` which returns a const reference to it.

I'm unsure though in regards to whether or not to make it private. Currently is protected, making it private would then make it a breaking change. 

This resolves #224  then.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -2530,7 +2530,7 @@ namespace jwt {
 	class decoded_jwt : public header<json_traits>, public payload<json_traits> {
 	protected:
 		/// Unmodifed token, as passed to constructor
-		const typename json_traits::string_type token;
+		typename json_traits::string_type token;
 		/// Header part decoded from base64
 		typename json_traits::string_type header;
 		/// Unmodified header part in base64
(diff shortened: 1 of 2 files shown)


Thalhammer:
While I am in no way against making it private (you are not really supposed to use it and modifying it wasnt possible anyway) I don't really see a benefit in doing so 🤔

Ka0o0:
Hi Dominik,
thank you for the quick response. Yes true. I was just considering this because the member was const but I also think it's fine to keep it protected as also the other fields are protected. Then I guess this PR is ready for merge.

Thalhammer:
Thanks for finding and fixing this. Great work. We might reconsider making this stuff private for the next major breaking release, but I'd like to avoid breaking changes (regardless how minor they are) unless we have to or there are already breaks anyway.

PR #231, opened 2022-05-19
Allow to retrieve list of claims

Ka0o0:
Currently, it is not possible to iterate over the list of claims of a JWK. With this commit, the JWK gains a new method that exposes the underlying claims so that the callee can do some processing with it. This might be useful, if someone wants to copy all claims to a new object.


prince-chrismc:
Would it be possible to add a test case please? 🙏

prince-chrismc:
Just a note to myself this is how its done for JWT payload and header claims

prince-chrismc:
The method has previously been untested for other components 🙈

PR #235, opened 2022-06-13
Update SSL libraries to latest versions

prince-chrismc:
### New List

image

UPDATE: everything is green

### Other changes

- Fixed a bad URL
- Fix badge action not reading job status 🤷

### DIFF
--- a/.github/actions/badge/action.yml
+++ b/.github/actions/badge/action.yml
@@ -5,7 +5,7 @@ inputs:
     description: The subfolder where to group the badges
     required: true
   label:
-    description: The label to you in the badge (this should be unqie for each badge in a category)
+    description: The label to you in the badge (this should be unique for each badge in a category)
     required: true
   github_token:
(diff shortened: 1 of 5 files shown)


Thalhammer:
Seems like one of the tests is failing cause the `ECDSA_SIG` isn't public anymore.
However there seems to be `ECDSA_SIG_get0_r` and `ECDSA_SIG_get0_s` which does exactly the same thing.
We need to make sure its available in everything we support thought so we don't introduce regressions for older libraries.

prince-chrismc:
> We might wanna go with the `ECDSA_SIG_get0_r` code

Sadly it only came out in the last version... 
https://github.com/openssl/openssl/commit/0396401d1c3fd65487116b0623e634b65bf28670
so it's does not look like too much of a gain

***

Thought looking into it we already are using this method with `ECDSA_SIG_get0` L1168 in jwt.h so it seems like our past selves were just as smart 😆

PR #240, opened 2022-09-13
Fix as_date()'s std::bad_cast with decimal value

Prosperoh:
If the timestamp value for the date is a decimal number, round it to the closest integer. This allows date claims to have a decimal value, though it will be rounded up to the closest second.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -23,6 +23,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <cmath>
 #include <codecvt>
 #include <functional>
 #include <iterator>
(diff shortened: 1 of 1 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -2253,11 +2253,25 @@ namespace jwt {
...
+			std::time_t timestamp;
+			if (get_type() == json::type::number)
+			{
+				timestamp = std::round(as_number());
+			}
+			else
+			{
+				timestamp = as_int();
+			}
+			return std::chrono::system_clock::from_time_t(timestamp);
```
Thalhammer, line 2273: `return std::chrono::system_clock::from_time_t(get_type() == json::type::number ? std::round(as_number()) : as_int());

Prosperoh: Updated.

Thalhammer (approved):

PR #251, opened 2022-10-01
Allow generic access to JWT header and payload

### LINKED ISSUE #248: Decode JWT Token (Boost)
### What's your question?

How to decode JWT token (HS256) with Boost traits?

### Additional Context

Hello. 
I have a token generated by https://jwt.io but I have a problem - Exception: Invalid input: not within alphabet.
How to fix it?
**Encoded:**
```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
```
**Header:**
```
{
  "alg": "HS256",
  "typ": "JWT"
}
```
**Payload:**
```
{
  "sub": "1234567890",
  "name": "John Doe",
  "iat": 1516239022
}
```
**VERIFY SIGNATURE:**
```
HMACSHA256(
  base64UrlEncode(header) + "." +
  base64UrlEncode(payload), 
your-256-bit-secret
)
```
**Code:**
```
#include <boost/json/src.hpp>
#include <jwt-cpp/traits/boost-json/traits.h>

using traits = jwt::traits::boost_json;
using claim = jwt::basic_claim<traits>;

void jwt_token(const std::string& token)
{
    const auto decoded = jwt::decode<traits>(token);
    std::cout << "decoded token: " << decoded.get_algorithm() << std::endl;
}
```
**Error log:**
```
token: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
Exception: Invalid input: not within alphabet
```

prince-chrismc:
closes #248 by allow direct access to the JSON. This section of code is only used to debug so it should be better to not have the hassle of the claim wrapping the data

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ int main() {
     std::string token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXUyJ9.eyJpc3MiOiJhdXRoMCJ9.AbIJTDMFc7yUa5MhvcP03nJPyCPzZtQcGEp-zWfOkEE";
     auto decoded = jwt::decode(token);
 
-    for(auto& e : decoded.get_payload_claims())
+    for(auto& e : decoded.get_payload_json())
         std::cout << e.first << " = " << e.second << std::endl;
 }
 ```
--- a/example/conan/main.cpp
+++ b/example/conan/main.cpp
@@ -12,6 +12,6 @@ int main() {
 
 	auto decoded = jwt::decode(token);
 
-	for (auto& e : decoded.get_payload_claims())
+	for (auto& e : decoded.get_payload_json())
 		std::cout << e.first << " = " << e.second << std::endl;
 }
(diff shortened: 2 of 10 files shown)


Thalhammer (approved):
This is a pretty severe API break though, but given I wanna get rid of `claim` in the long term anyway I am fine with that.

prince-chrismc:
Yep! Sever > surprise 

This way it's very obvious the need to update the code... instead of not having access to old memeber functions.

PR #252, opened 2022-10-01
Define a "jwt base class" for exceptions

### LINKED ISSUE #215: [bug] Use new common base class for all exceptions thrown by jwt-cpp
This was initially described in https://github.com/Thalhammer/jwt-cpp/issues/212. Please add new class derived from `std::system_error`, and derive all jwt-cpp exceptions from it. This would allow to write bigger block of code, and at the end add few catch sections for different kinds of errors (e.g. network error, jwt validation error, etc.). You also could store some extra information in additional fields of exception class, e.g. name of claim which did not pass validation.

**Desktop:**
- OS: RedHat 8
- Compiler g++ 10.3.1
- Version jwt-cpp v0.6.0

prince-chrismc:
closes #215

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -90,19 +90,23 @@ namespace jwt {
 	 * \brief Everything related to error codes issued by the library
 	 */
 	namespace error {
-		struct signature_verification_exception : public std::system_error {
+		struct exception : public std::exception {
+			using exception::exception;
+		}
+
+		struct signature_verification_exception : exception, public std::system_error {
 			using system_error::system_error;
 		};
-		struct signature_generation_exception : public std::system_error {
+		struct signature_generation_exception : exception, public std::system_error {
 			using system_error::system_error;
 		};
-		struct rsa_exception : public std::system_error {
+		struct rsa_exception : exception, public std::system_error {
 			using system_error::system_error;
 		};
-		struct ecdsa_exception : public std::system_error {
+		struct ecdsa_exception : exception, public std::system_error {
 			using system_error::system_error;
 		};
-		struct token_verification_exception : public std::system_error {
+		struct token_verification_exception : exception, public std::system_error {
 			using system_error::system_error;
 		};
 		/**
@@ -362,15 +366,8 @@ namespace jwt {
 			}
 		}
 	} // namespace error
-
-	// FIXME: Remove
-	// Keep backward compat at least for a couple of revisions
-	using error::ecdsa_exception;
-	using error::rsa_exception;
-	using error::signature_generation_exception;
-	using error::signature_verification_exception;
-	using error::token_verification_exception;
 } // namespace jwt
+
 namespace std {
 	template<>
 	struct is_error_code_enum<jwt::error::rsa_error> : true_type {};
@@ -383,6 +380,7 @@ namespace std {
 	template<>
 	struct is_error_code_enum<jwt::error::token_verification_error> : true_type {};
 } // namespace std
+
 namespace jwt {
 	/**
 	 * \brief A collection for working with certificates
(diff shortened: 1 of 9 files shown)


prince-chrismc:
There's actually a lot of these left 

image

include/jwt-cpp/jwt.h
```diff
@@ -90,19 +90,26 @@ namespace jwt {
...
 	namespace error {
-		struct signature_verification_exception : public std::system_error {
+		/**
+		 * \brief Generic base class for all JWT specific exceptions
+		 */
+		struct exception : public std::exception {
+			using std::exception::exception;
+		};
+
+		struct signature_verification_exception : exception, public std::system_error {
```
Thalhammer, line 100: This feels really weird cause now signature_verification_exception inherits from std::exception twice.
Could we get away with inheriting `exception` from `std::system_error` (which inherits `std::runtime_error`, which in turn inherits `std::exception`) ?

include/jwt-cpp/jwt.h
```diff
@@ -2316,13 +2321,13 @@ namespace jwt {
 		/**
 		 * Attempt to parse JSON was unsuccessful
 		 */
-		struct invalid_json_exception : public std::runtime_error {
+		struct invalid_json_exception : exception, public std::runtime_error {
```
Thalhammer, line 2324: See above, given that we dont actually set a dynamic error message, we should probably add a "invalid_json" error code and stick that in.

Thalhammer (changes requested):

prince-chrismc:
I think we need a more clear goal for this one. I was not expecting it to be soo big 🙈 

On the server side if you only do this

```cpp
try { 
     process_jwt(token);
catch (jwt_exception e) {
     //... handling 
}
```

How can you know to return 401 vs 403 😕 there's a difference for the token being malformed and not granting permission 🤔

Thalhammer:
> there's a difference for the token being malformed and not granting permission

Wouldn't this normally be handled in two different places ?
First you check if the token is a valid token (i.e. not malformed, not expired, etc.) then you check if the token (which you can now trust) actually allows access to the resource at hand. At least in my code bases this doesn't even happen in the same place, the first check is inside a middleware whereas the second is in the individual handler.

Normally the difference between multiple error groups is handled by `std::error_category`, if the token is malformed you'd get `signature_verification_error_category()` as the category, whereas a expired token (or one missing claims,etc) would give you a `token_verification_error_category()` error. So your handling (assuming you check the resource access inside the verifier) would look something like this:
```cpp
try { 
     process_jwt(token);
catch (jwt_exception e) {
    if(e.code().category() == signature_verification_error_category())
         return 401;
    if(e.code().category() == token_verification_error_category())
         return 403;
    return 500;
}
```
You could also dig down into the actual error code, e.g. to distinguish `token_expired` and `audience_missmatch`.

Regardless, the fact that all the exception types now inherit std::exception twice definitly feels wrong and will almost certainly cause issues. What we could do is have a `jwt::exception` inherit `std::system_error` (and thus `std::runtime_error` and `std::exception`) and then inherit something like `invalid_json_exception` from `jwt::exception` so you can still catch those individually instead of checking the error category if you want.

prince-chrismc:
Tomorrow while I travel, I will look at this again... I agree this isn't up to par. Theres 130+ exceptions that are from the STL... some are for sure traits but others I have no clue...

Once we know what the exceptions are doing we can figure out a few user stories. 

Come up with a plan to improve this.

prince-chrismc:
I mean we are using them a lot 🤷 despite the other ways we are also using

prince-chrismc:
Excluding the traits... there's 5 left

image

- 2 are from the evp key ref
- 2 are invalid tokens
- last is claim compare in verifier

PR #256, opened 2022-10-04
Add helpers to some crypto APIs that vary by version 

prince-chrismc:
the idea here is to dry up the code since there was a bunch of duplication

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -747,7 +747,7 @@ namespace jwt {
 		inline evp_pkey_handle load_public_ec_key_from_string(const std::string& key, const std::string& password,
 															  std::error_code& ec) {
 			ec.clear();
-			pubkey_bio = make_mem_buf_bio();
+			auto pubkey_bio = make_mem_buf_bio();
 			if (!pubkey_bio) {
 				ec = error::ecdsa_error::create_mem_bio_failed;
 				return {};
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -470,40 +470,28 @@ namespace jwt {
 #endif
 		};
 
-		struct unique_bio_ptr : std::unique_ptr<BIO, decltype(&BIO_free_all)> {
-			unique_bio_ptr() : std::unique_ptr<BIO, decltype(&BIO_free_all)>(BIO_new(BIO_s_mem()), BIO_free_all) {}
-			unique_bio_ptr(const std::string& data)
-				: std::unique_ptr<BIO, decltype(&BIO_free_all)>(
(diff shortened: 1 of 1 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -469,24 +469,54 @@ namespace jwt {
...
+#else
+					  BIO_new_mem_buf(data.data(), static_cast<int>(data.size())), BIO_free_all
+#endif
+				  ) {
+			}
+
+			unique_bio_ptr(const unique_bio_ptr&) = delete;
+			unique_bio_ptr(unique_bio_ptr&&) noexcept = default;
+			~unique_bio_ptr() noexcept = default;
+		};
```
Thalhammer, line 488: I am not really a fan of inheriting `std` classes, especially since in this case they barely do anything and the name does not imply that it creates a mem bio. Maybe something like
```cpp
std::unique_ptr<BIO, decltype(&BIO_free_all)> make_mem_buf_bio(const std::string& data) {
	return std::unique_ptr<BIO, decltype(&BIO_free_all)>(
#if OPENSSL_VERSION_NUMBER <= 0x10100003L
		BIO_new_mem_buf(const_cast<char*>(data.data()), static_cast<int>(data.size())), BIO_free_all
#else
		BIO_new_mem_buf(data.data(), static_cast<int>(data.size())), BIO_free_all
#endif
	);
}
```
Is a better fit ?

include/jwt-cpp/jwt.h
```diff
@@ -469,24 +469,54 @@ namespace jwt {
...
+			std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_free)>
+#endif
+		{
+			unique_evp_md_ctx_ptr()
+				:
+#ifdef JWT_OPENSSL_1_0_0
+				  std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_destroy)>(EVP_MD_CTX_create(), &EVP_MD_CTX_destroy)
+#else
+				  std::unique_ptr<EVP_MD_CTX, decltype(&EVP_MD_CTX_free)>(EVP_MD_CTX_new(), EVP_MD_CTX_free)
+#endif
```
Thalhammer, line 503: See above. `EVP_MD_CTX_destroy` and `EVP_MD_CTX_free` have the same signature, so we can just use `void(*)(EVP_MD_CTX*)` as the destructor type (which is what both decltypes would resolve to anyway. Also in the `EVP_MD_CTX_free` branch theres a missing `&` at the end.

Thalhammer (changes requested):

prince-chrismc:
I like it, saves 20 lines + there are 7 fewer `#ifdef` in the code which is check my goal

Thalhammer (approved):

Issue #243, opened 2022-09-16
NuGet package for jwt-cpp


diogo-strube:
### What's your question?

Can I create and publish a NuGet package?

### Additional Context

Hi! I have been using this library and loving it!
The caveat is my current project heavily uses NuGet to organize dependencies... would it be ok if I package jwt-cpp and publish it to [nuget.org](https://www.nuget.org/)?
Of course, the package metadata would point to this repo, the existing readme/license, and have @Thalhammer as owner.

prince-chrismc:
Yes 🙏 if you want to contribute an [Action[(https://github.com/marketplace/actions/publish-nuget) or an example that would be amazing ❤️ 

Nuget is only <9% from this year's isocpp study but it would be great to reach more developers!

Thalhammer:
> Can I create and publish a NuGet package?

Of course, we'd be happy if you want to create one. I didn't even know nuget was a thing for C++, I always assumed its only for C# and the C++ equivalent is vcpkg (which we do support).

> Nuget is only <9% from this year's isocpp study

Hunter is probably not even in there, yet we support it :)

diogo-strube:
Thank you @prince-chrismc and @Thalhammer ❤️
I will be happy to contribute with a GitHub Action 😀

`Nuget is only <9% from this year's isocpp study but it would be great to reach more developers!`
Could you share a link to this study? I promise I am not lazy, hehe, I tried searching for it only but couldn't find it 😅

`I didn't even know nuget was a thing for C++, I always assumed its only for C# and the C++ equivalent is vcpkg (which we do support).`
NuGet definitely has a strong history with C#! But for C++ devs using Visual Studio as an IDE, NuGet packages are an excellent option for a "pure clone & build" experience. The caveat is that vcpkg [integration with VS](https://vcpkg.io/en/docs/users/buildsystems/msbuild-integration.html) (.targets and .props files) that allows "auto install" requires vcpkg as a submodule, so you can still have someone confused with errors after cloning the repo without git `--recurse-submodules`. So, with NuGet Pkgs I noticed less confusion across fellow developers.

prince-chrismc:
Shameless plug but we also have a Conan package if you need something more C++ friendly 😉 

> Could you share a link to this study? I promise I am not lazy, hehe, I tried searching for it online but couldn't find it 😅

https://isocpp.org/files/papers/CppDevSurvey-2022-summary.pdf

Question 7 -- it's 9.3%

[...]

PR #208, opened 2022-02-11
🏷️ Set version to 0.6.0

prince-chrismc:
Everything seems to be inplace 🤞

PR #262, opened 2022-10-12
NuGet Pkg for v0.6.0 using GitHub Actions

diogo-strube:
This PR is related to discussion #243 and is a 3 file contribution:
- nuget\jwt-cpp.nuspec contains all the descriptions of the NuGet package, including authors, and the path of files to be copied.
- nuget\jwt-cpp.targets automatically set up the "additional include directories" when the package is installed.
- .github\workflows\nuget.yml contains the GitHub Action to pack and pub the NuGet package.

Some context on the current approach:
- As the version is already tracked by committed files (as shown in #208), I left the version explicit in the .nuspec file (instead of using an environment variable or secret).
- Publishing a NuGet pkg (to nugte.org) requires a token, the same was configured as a secret ${{ secrets.nuget_api_key }} in .github\workflows\nuget.yml.

### DIFF
--- a/.github/workflows/nuget.yml
+++ b/.github/workflows/nuget.yml
@@ -1,0 +1,37 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Nuget
+
+# Controls when the workflow will run
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v3
+        with:
+          ref: 'v${{ secrets.current_version }}'
+      #   repository: 'diogo-strube/jwt-cpp'
+      
+      # Deploy NuGet so we can call the pack and push cmds
+      - name: Setup NuGet.exe for use with actions
+        uses: NuGet/setup-nuget@v1.0.7
+        with:
+          nuget-api-key: ${{ secrets.nuget_api_key }}
+          
+      # Package the nusepc file
+      - name: Create NuGet pkg
+        run: nuget pack jwt-cpp.nuspec
+        
+      # Publish nuget pkg
+      - name: Publish NuGet pkg
+        run: nuget push *.nupkg -Source 'https://api.nuget.org/v3/index.json' -Version '${{ secrets.current_version }}'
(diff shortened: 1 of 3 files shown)


.github/workflows/nuget.yml
```diff
@@ -0,0 +1,39 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Nuget
+
+# Controls when the workflow will run
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
prince-chrismc, line 8: Have you considered using https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release so that it will get pushed automatically?

I do like manual so we can test but less is more 😉

.github/workflows/nuget.yml
```diff
@@ -0,0 +1,39 @@
...
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v3
+        with:
+          ref: 'v0.6.0'
+      
+      # Deploy NuGet so we can call the pack and push cmds
+      - name: Setup NuGet.exe for use with actions
+        uses: NuGet/setup-nuget@v1.0.7
+        with:
+          nuget-api-key: ${{ secrets.nuget_api_key }}
```
prince-chrismc, line 28: Do we need access to the Nuget package to make an API token? I made an account with the same user name if you don't mind :)

.github/workflows/nuget.yml
Thalhammer: I think manual is fine while we test, we can switch it later once we are more familiar with how it works.

.github/workflows/nuget.yml
Thalhammer: Yes, he needs to add one of us (or ideally both) as a owner for the package and we then need to create a secret [here](https://github.com/Thalhammer/jwt-cpp/settings/secrets/actions) with the push token.

Thalhammer (approved):

.github/workflows/nuget.yml
prince-chrismc: You are the only one who can add secrets AFAIK, I dont have access to the settings page where they can be added

(my evil genius plan to steal power is working muahahah 😈 )

Thalhammer: > I dont have access to the settings page where they can be added

Guess I need to add them myself then :smiling_imp:

diogo-strube: Added you both access as owners directly via NuGet.org so you can manage the package even before we publish the next one:
image
And the next commit will have you both added as owners in the .nuspec file, together with the version updated to a test one so you can experiment with running the workflow with your own token :)

diogo-strube:
@Thalhammer and @prince-chrismc, with the latest commit you should be ready to configure the secret and try running the workflow, it will generate a pre-release package "-dev0.2" that we can do a final validation before publishing the release package.

Thalhammer:
@prince-chrismc Can you create a token for nuget ?
I can't cause I am not a co owner of the package but you seem to be now.

prince-chrismc:
should be in you inbox :)

PR #263, opened 2022-10-12
Reserve & append string in json_traits sign

JaredTherriault:
iOS crashes when calling sign twice in the same instance - this fix reserves the string memory ahead of time to prevent this.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -2891,14 +2891,22 @@ namespace jwt {
 			typename json_traits::object_type obj_header = header_claims;
 			if (header_claims.count("alg") == 0) obj_header["alg"] = typename json_traits::value_type(algo.name());
 
-			const auto header = encode(json_traits::serialize(typename json_traits::value_type(obj_header)));
-			const auto payload = encode(json_traits::serialize(typename json_traits::value_type(payload_claims)));
-			const auto token = header + "." + payload;
+			const std::string header = encode(json_traits::serialize(typename json_traits::value_type(obj_header)));
+			const std::string payload = encode(json_traits::serialize(typename json_traits::value_type(payload_claims)));
+			std::string token;
+			token.reserve(header.length() + 1 + payload.length());
+			token.append(header).append(".").append(payload);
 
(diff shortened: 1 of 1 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -2891,14 +2891,22 @@ namespace jwt {
 			typename json_traits::object_type obj_header = header_claims;
 			if (header_claims.count("alg") == 0) obj_header["alg"] = typename json_traits::value_type(algo.name());
 
-			const auto header = encode(json_traits::serialize(typename json_traits::value_type(obj_header)));
-			const auto payload = encode(json_traits::serialize(typename json_traits::value_type(payload_claims)));
-			const auto token = header + "." + payload;
+			const std::string header = encode(json_traits::serialize(typename json_traits::value_type(obj_header)));
```
prince-chrismc, line 2894: Ca you please use auto? People are able to provide their own methods for other types

prince-chrismc (commented):
Would it be possible to add a test case for this?

I mean calling sign sign should work but I am surprised it's iOS spefic. I'd love to see an action to check that platform

Thalhammer:
No offense, but this seems like a work around for a problem in your code/compiler/stl. If I understood it correctly it boils down to:
```cpp
std::string str3 = str1 + str2;
```
vs
```
std::string str3;
str3.reserve(str1.size() + str2.size());
str3.append(str2);
str3.append(str2);
```
which according to the standard should behave exactly the same. The later might be slightly more efficient as it prevents a realloc of the buffer, but that shouldn't cause crashes. I feel like theres some sort of race condition or something like that, thats now hidden because the resize causes the timing to shift slightly enough to not cause issues anymore.

Can you provide more details on where it crashes (e.g. a stack trace) ?

OpenSSL (and by extend jwt-cpp) is not threadsafe by default on most platforms and jwt-cpp makes no attempt in being it either, so if you reuse the same instance make sure not to do so from two threads at the same time.

Thalhammer:
@JaredTherriault Any update ?

prince-chrismc:
I'll close this for now, if you have more information feel free to reopen

PR #268, opened 2022-10-31
:arrow_up: Upgrade various ci packages being deprecated soon

Thalhammer:


prince-chrismc:
We should add depenabot for the github actions 🤔

Thalhammer:
I don't really like dependabot because it creates tons of pr's for simple things that can be done in a single commit or along with a different change and are not urgent.

prince-chrismc:
That's fair

PR #270, opened 2022-11-03
casting der_signature.length() to 'long' to silence compile warning

gizmocuz:


prince-chrismc (approved):
Same as the other place we do this https://github.com/Thalhammer/jwt-cpp/blob/d403e47f854e6ed7fc19bc65a3e155d94ce5fbed/include/jwt-cpp/jwt.h#L1218

PR #272, opened 2022-12-05
Rough attempt at a `openssl_base64_decode`

### LINKED ISSUE #89: Move to OpenSSL base64
We relly on openssl pretty heavily already, so why not use its base64 support instead of the one I originaly wrote myself.
I originaly refused to use it, because it seems hard/complicated to use and I already had a header only implementation in a personal project which I just had to drop in.

However I recently found out there seems to be a way easier approach to use it now:
https://stackoverflow.com/a/60580965/7992576

Might be worth checking out.

prince-chrismc:
resolves #89

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -468,6 +468,10 @@ namespace jwt {
 			return std::unique_ptr<BIO, decltype(&BIO_free_all)>(BIO_new(BIO_s_mem()), BIO_free_all);
 		}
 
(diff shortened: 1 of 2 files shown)


prince-chrismc:
Might stick some `GCC_VERSION >= 5` in there if it's the only failure

Thalhammer:
> Might stick some `GCC_VERSION >= 5` in there if it's the only failure

Seems like the wolfSSL Tests fail as well, which is probably a bigger impact than gcc 5.

prince-chrismc:
I am pretty sure the wolfssl handles base64 differently or generates different certs.

Traveling for the holidays but hopefully I'll get to look into it

PR #275, opened 2023-01-15
std::string_view support

sjanel:
If the json library supports `std::string_view`, allow `jwt-cpp` take builder parameters as `std::string_view` and algorithm constructors as well.

Also small optimizations in some methods of `jwt.h` to make sure only one object is returned so that RVO can be used.

PR is ready for review. Only the coverage is failing, I will check why.

### DIFF
--- a/include/jwt-cpp/base.h
+++ b/include/jwt-cpp/base.h
@@ -2,11 +2,11 @@
 #define JWT_CPP_BASE_H
 
 #include <algorithm>
-#include <array>
 #include <cstdint>
 #include <stdexcept>
 #include <string>
-#include <vector>
+
+#include "string_types.h"
 
 #ifdef __has_cpp_attribute
 #if __has_cpp_attribute(fallthrough)
@@ -16,6 +16,11 @@
 
 #ifndef JWT_FALLTHROUGH
 #define JWT_FALLTHROUGH
+#endif
+
+#ifndef JWT_HAS_STRING_VIEW
+#include <array>
+#include <cstring>
 #endif
 
 namespace jwt {
@@ -30,19 +35,31 @@ namespace jwt {
 		 * base64-encoded as per [Section 4 of RFC4648](https://datatracker.ietf.org/doc/html/rfc4648#section-4)
 		 */
 		struct base64 {
+
+#define JWT_BASE_ALPHABET                                                                                              \
+	'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', \
+		'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's',  \
+		't', 'u', 'v', 'w', 'x', 'y', 'z', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9'
+
+#ifdef JWT_HAS_STRING_VIEW
+			// From C++17 it's perfectly fine to have inline static variables. No ODR violation in this case.
+			static constexpr char kData[]{JWT_BASE_ALPHABET, '+', '/'};
+
+			static constexpr std::string_view kFill[]{"="};
+#else
+			// For pre C++17 standards, we need to use a method
 			static const std::array<char, 64>& data() {
-				static constexpr std::array<char, 64> data{
-					{'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P',
-					 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f',
-					 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
-					 'w', 'x', 'y', 'z', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '+', '/'}};
-				return data;
+				static constexpr std::array<char, 64> kData{{JWT_BASE_ALPHABET, '+', '/'}};
+				return kData;
 			}
-			static const std::string& fill() {
-				static std::string fill{"="};
-				return fill;
+
+			static const std::array<const char*, 1>& fill() {
+				static constexpr std::array<const char*, 1> kFill{"="};
+				return kFill;
 			}
+#endif
 		};
+
 		/**
 		 * \brief valid list of character when working with [Base64URL](https://tools.ietf.org/html/rfc4648#section-5)
 		 *
@@ -53,18 +70,24 @@ namespace jwt {
 		 * > [Section 5 of RFC 4648 RFC4648](https://tools.ietf.org/html/rfc4648#section-5), with all trailing '=' characters omitted
 		 */
 		struct base64url {
+
+#ifdef JWT_HAS_STRING_VIEW
+			static constexpr char kData[]{JWT_BASE_ALPHABET, '-', '_'};
+
+			static constexpr std::string_view kFill[]{"%3d"};
+#else
+			// For pre C++17 standards, we need to use a method
 			static const std::array<char, 64>& data() {
-				static constexpr std::array<char, 64> data{
-					{'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P',
-					 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f',
-					 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
-					 'w', 'x', 'y', 'z', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '-', '_'}};
-				return data;
+				static constexpr std::array<char, 64> kData{{JWT_BASE_ALPHABET, '-', '_'}};
+				return kData;
 			}
-			static const std::string& fill() {
-				static std::string fill{"%3d"};
-				return fill;
+
+			static const std::array<const char*, 1>& fill() {
+				static constexpr std::array<const char*, 1> kFill{"%3d"};
+				return kFill;
 			}
+
+#endif
 		};
 		namespace helper {
 			/**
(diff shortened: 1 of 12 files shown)


include/picojson/picojson.h
```diff
@@ -117,6 +119,14 @@ extern "C" {
 
 namespace picojson {
```
prince-chrismc, line 120: This is a thirdparty library and I am not sure changing the source code here is going to help us move away from it since it's no longer maintained

tests/BaseTest.cpp
```diff
@@ -1,110 +1,119 @@
...
-	ASSERT_EQ("MTIzNA", jwt::base::trim<jwt::alphabet::base64>("MTIzNA=="));
-}
-
-TEST(BaseTest, Base64TrimURL) {
-	ASSERT_EQ("MQ", jwt::base::trim<jwt::alphabet::base64url>("MQ%3d%3d"));
-	ASSERT_EQ("MTI", jwt::base::trim<jwt::alphabet::base64url>("MTI%3d"));
-	ASSERT_EQ("MTIz", jwt::base::trim<jwt::alphabet::base64url>("MTIz"));
-	ASSERT_EQ("MTIzNA", jwt::base::trim<jwt::alphabet::base64url>("MTIzNA%3d%3d"));
-}
+namespace jwt {
```
prince-chrismc, line 4: Would you mind removing these namespaces it introduces a lot of change and it more difficult to see whats different

include/jwt-cpp/traits/boost-json/traits.h
```diff
@@ -67,7 +67,8 @@ namespace jwt {
 				return val.get_double();
 			}
 
-			static bool parse(value_type& val, string_type str) {
+			template<class string_t>
```
prince-chrismc, line 70: should this not be specialized based on what the library supports?

include/jwt-cpp/traits/nlohmann-json/traits.h
```diff
@@ -64,7 +64,8 @@ namespace jwt {
 				return val.get<double>();
 			}
 
-			static bool parse(json& val, std::string str) {
+			template<class string_t>
+			static bool parse(json& val, const string_t& str) {
```
prince-chrismc, line 68: Would it not be more ideal to specialize these so users do not make mistakes

https://github.com/nlohmann/json/blob/6af826d0bdb55e4b69e3ad817576745335f243ca/README.md?plain=1#L1640

There could be one for both types?

include/jwt-cpp/jwt.h
```diff
@@ -494,38 +496,39 @@ namespace jwt {
...
-													std::error_code& ec) {
+		inline std::string extract_pubkey_from_cert(string_view certstr, const std::string& pw, std::error_code& ec) {
 			ec.clear();
 			auto certbio = make_mem_buf_bio(certstr);
 			auto keybio = make_mem_buf_bio();
+			std::string res;
 			if (!certbio || !keybio) {
 				ec = error::rsa_error::create_mem_bio_failed;
-				return {};
+				return res;
```
prince-chrismc, line 506: Have you checked how this impacts compiler optimizations? Last I checked the compilers are better at implementing copy elision with r-values

include/jwt-cpp/jwt.h
```diff
@@ -675,7 +677,7 @@ namespace jwt {
 
 			evp_pkey_handle pkey(PEM_read_bio_PUBKEY(
 				pubkey_bio.get(), nullptr, nullptr,
-				(void*)password.data())); // NOLINT(google-readability-casting) requires `const_cast`
+				(void*)password.c_str())); // NOLINT(google-readability-casting) requires `const_cast`
```
prince-chrismc, line 680: What is the reason for this?

include/jwt-cpp/string_types.h
```diff
@@ -0,0 +1,24 @@
...
+#define JWT_HAS_STRING_VIEW
+
+	using string_constant = std::string_view;
+	using string_view = std::string_view;
+
+#else
+
+#include <string>
+
+	using string_constant = std::string;
```
prince-chrismc, line 18: This name seems misleading 🤔 a non-const constant string?

tests/BaseTest.cpp
```diff
@@ -1,110 +1,119 @@
...
+
+	namespace {
+		base::details::padding count_padding(string_view base, std::initializer_list<std::string> fills) {
```
prince-chrismc, line 7: Should this not be in base details with the other implementation since users would need it?

include/jwt-cpp/base.h
```diff
@@ -81,15 +76,15 @@ namespace jwt {
...
-					static std::initializer_list<std::string> fill{"%3D", "%3d"};
-					return fill;
-				}
+				static std::array<string_constant, 2> fill() { return std::array<string_constant, 2>{"%3D", "%3d"}; }
 			};
 		} // namespace helper
 
 		inline uint32_t index(const std::array<char, 64>& alphabet, char symbol) {
-			auto itr = std::find_if(alphabet.cbegin(), alphabet.cend(), [symbol](char c) { return c == symbol; });
+			if (symbol >= 'A' && symbol <= 'Z') { return static_cast<uint32_t>(symbol - 'A'); }
```
prince-chrismc, line 84: Have you checked the performance hit for all these branches?

include/jwt-cpp/base.h
```diff
@@ -116,30 +111,31 @@ namespace jwt {
 				}
 			};
 
-			inline padding count_padding(const std::string& base, const std::vector<std::string>& fills) {
-				for (const auto& fill : fills) {
-					if (base.size() < fill.size()) continue;
+			template<class StrInputIt>
+			padding count_padding(string_view base, StrInputIt fillStart, StrInputIt fillEnd) {
```
prince-chrismc, line 115: Could you please change these to snake_case to match the rest of the code base?

prince-chrismc (commented):
I generally understand the motivation behind this contribution but I would generally suggest opening an issue to discuss ideas this large, tip for the future :)

There are a handful of changes which are a little unclear what the motivation so I left some comments during an initial pass.

Overall the code looks really good, my comments are for existing users who might want to update as well as some performance consideration :)

include/jwt-cpp/base.h
sjanel: Yes sure.

sjanel:
> I generally understand the motivation behind this contribution but I would generally suggest opening an issue to discuss ideas this large, tip for the future :)
> 
> There are a handful of changes which are a little unclear what the motivation so I left some comments during an initial pass.
> 
> Overall the code looks really good, my comments are for existing users who might want to update as well as some performance consideration :)

Thanks for the comments, I was quite busy these days, but I will try to come back on my PR, I did not forget it ;)
I will try to split it as well. I think I can extract some stuff in `base.h` first.

include/jwt-cpp/string_types.h
sjanel: The name is probably not the best one. It should only be used for definition of compile time constant strings. In C++17 we can (and it's the best solution) use `std::string_view`, but older C++ versions don't have it so the best we can do is to default on `std::string`.

These types are the building blocks of this PR, and makes it possible for pre C++17 compilers to still compile the library. Do you have a better name in mind ? For instance, `static_string` ?

include/jwt-cpp/traits/boost-json/traits.h
sjanel: `json::parse` accepts a templated type. IMHO it's safe to also take any string type.

tests/BaseTest.cpp
sjanel: I like them. Unit tests of this lib should be enclosed in the same namespace as the modules they test, hence jwt. Will make a PR with only the namespace changes to make the differences clearer here.

include/jwt-cpp/base.h
sjanel: Here it is: https://quick-bench.com/q/8u7o4DL5yBu256Rzgu_KLi2Z4JA

include/jwt-cpp/jwt.h
sjanel: Having only one returned variable enables use of RVO (in the modified code). I believe that new code may be slightly more efficient, although I did not check. The impacts should be minor though, it's more a matter of code style in the end. If it's an issue for you, I can revert these modifications.

include/jwt-cpp/base.h
Thalhammer: Keep in mind, that the new code is not a direct equivalent to the original though. It relies on the fact that the first 62 chars of the alphabet are "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789" which effectively demotes the alphabet type to a 2 char field. I am not sure if this is a smart thing to do, it doesn't feel right to provide a user hook to change the whole alphabet just to ignore 96% of it. In the best case this leads to inconvenience, in the worst to bugs.

include/picojson/picojson.h
Thalhammer: If you need to modify the json library for it to work with jwt-cpp then we can't do so much string_view. jwt-cpp aims to be as agnostic as possible to json/ssl libraries.

Thalhammer (changes requested):

sjanel (commented):
- Reverted changes based on `std::string res` as suggested in code review comments
- alg name as `std::string` passed by copy then `std::move`

include/jwt-cpp/base.h
```diff
@@ -108,39 +140,44 @@ namespace jwt {
...
 				return {};
 			}
 
-			inline std::string encode(const std::string& bin, const std::array<char, 64>& alphabet,
-									  const std::string& fill) {
+			inline std::string encode(string_view bin, const char* alphabet, string_view fill) {
 				size_t size = bin.size();
 				std::string res;
 
+				res.reserve((4UL * size) / 3UL);
```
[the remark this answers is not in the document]
Thalhammer: I am not referring to a failure in the sense that it creates an error, however it will not prevent multiple allocations. The formula assumes fill is a single character and will allocate accordingly. reserve however does not follow the usual growth factor, it will allocate exactly the capacity passed to it (assuming its larger than the current size) meaning if we later add the fill and its not a single char the string has to reallocate to fit the extra chars. What I am proposing is to add a small extra of 2*fill.size() so that the reserved size is guaranteed to be larger than everything it will get.

Thalhammer:
For some reason I could not reply to a bunch of your comments, so here we go.

> It's at least equivalent considering the alphabets it receives. Do we expect the alphabets to change ? As this file is meant to be internal the risk is close to 0 to introduce a bug here (and it's unit tested anyway).

Likely not, but its exposed to the public and someone might use it with a different alphabet. Its a user visible header and the base64 stuff is not in a details namespace which makes it kinda public api (even if its not advertised as one). If we potentially break peoples code, we should at least break it at the compile stage (I.e. change the types inside the alphabets and remove the - no superficial - first 62 chars) instead of silently changing behavior in edge cases.

> If it's no longer maintained, why is it a problem anyway as it should be removed from this repo ?

Mostly because it being there didn't hurt so far and removing it would likely break a lot of users codebase because its the default library. Also having one in there that does not rely on a third party json library makes it way easier to use (no extra dependencies).

sjanel:
Hello @Thalhammer  @prince-chrismc 

I answered your comments, not sure what is preventing the PR to be merged currently.

Also, it fixes following gcc 12 compiling error:
```
[build] include/jwt-cpp/base.h: In static member function ‘static const std::initializer_list<std::__cxx11::basic_string<char> >& jwt::alphabet::helper::base64url_percent_encoding::fill()’:
[build] include/jwt-cpp/base.h:86:100: error: ‘(((const std::__cxx11::basic_string<char>*)(& <temporary>)) != 0)’ is not a constant expression
[build]    86 |                                         static std::initializer_list<std::string> fill{"%3D", "%3d"};
```

and an attempt to fix it 'easily' is more complicated than expected.

Here with this PR and string_view it's much easier.

sjanel:
> Mostly because it being there didn't hurt so far and removing it would likely break a lot of users codebase because its the default library. Also having one in there that does not rely on a third party json library makes it way easier to use (no extra dependencies).

That's the reason why I fixed it instead of removing it. They are not complicated changes, and backward compatible with ancient compilers. If we cannot remove it, and cannot change it, how can we move on ?

prince-chrismc:
I am sincerely sorry, I've been meaning to follow up but I am currently between jobs and trying to move countries. It might be several more weeks before I have a chance to work from my computer so I'd really appreciate your patients.

ambiennt:
Is there any way this could get merged in soon?  I find that supporting a string view interface is very important to avoid allocation overheads for string-like types that the jwt-cpp API doesn't need to take ownership of (like passing the public/private key pairs when constructing a `jwt::algorithm` type).

PR #283, opened 2023-03-11
added convenience function for no base64 der to pem

dennisyakovlev:
I think a helper function for der to pem cert if original is not base64 encoded would nice.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -545,13 +545,10 @@ namespace jwt {
 		/**
 		 * \brief Convert the certificate provided as DER to PEM.
 		 *
-		 * \tparam Any is used to stop linker errors instead of using inline.
-		 *
 		 * \param cert_der_str 	String containing the certificate encoded as base64 DER
 		 * \param ec			error_code for error_detection (gets cleared if no error occures)
 		 */
-		template<typename Any = void>
-		std::string convert_der_to_pem(const std::string& cert_der_str, std::error_code& ec) {
+		inline std::string convert_der_to_pem(const std::string& cert_der_str, std::error_code& ec) {
 			ec.clear();
 
 			auto c_str = reinterpret_cast<const unsigned char*>(cert_der_str.c_str());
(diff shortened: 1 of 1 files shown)


prince-chrismc (approved):

include/jwt-cpp/jwt.h
```diff
@@ -543,28 +543,21 @@ namespace jwt {
...
-		 * \brief Convert the certificate provided as base64 DER to PEM.
-		 *
-		 * This is useful when using with JWKs as x5c claim is encoded as base64 DER. More info
-		 * (here)[https://tools.ietf.org/html/rfc7517#section-4.7]
+		 * \brief Convert the certificate provided as DER to PEM.
 		 *
-		 * \tparam Decode is callabled, taking a string_type and returns a string_type.
-		 * It should ensure the padding of the input and then base64 decode and return
-		 * the results.
+		 * \tparam Any is used to stop linker errors instead of using inline.
```
Thalhammer, line 548: Whats the reasoning behind using a dummy template arg instead of inline directly ?
Given the comment I suspect you had a reasoning for the decision, I am curious what that was.

include/jwt-cpp/jwt.h
```diff
@@ -586,6 +579,28 @@ namespace jwt {
...
+		 *
+		 * \param cert_base64_der_str 	String containing the certificate encoded as base64 DER
+		 * \param decode 				The function to decode the cert
+		 * \param ec					error_code for error_detection (gets cleared if no error occures)
+		 */
+		template<typename Decode>
+		std::string convert_base64_der_to_pem(const std::string& cert_base64_der_str, Decode decode,
+											  std::error_code& ec) {
+			ec.clear();
+			const auto decodedStr = decode(cert_base64_der_str);
```
Thalhammer, line 600: I am on a holy war to free the world from the burden of camelCase and as such the entire codebase uses snake_case (or no spacing). Both `decoded_str`, `decodedstr` and `decoded` are fine with me.

Thalhammer (commented):
Just some minor nitpicks, I am fine with it otherwise.

include/jwt-cpp/jwt.h
dennisyakovlev: Yeah to be honest not sure how that got there as I use `camel_case`.

include/jwt-cpp/jwt.h
dennisyakovlev: The two reasons which are mostly preference. Firstly, imo its more likely to be interpreted incorrectly by a reader. Secondly, an unnecessary `inline` can maybe screw with the compiler so Instead of taking the risk I use a substitute.

However, I think consistency should be prioritized.

prince-chrismc (approved):

PR #290, opened 2023-05-08
Fix compiler warning (clang, cxx17)

JonasBorchelt:
warnings: 

`'wstring_convert<std::codecvt_utf8<wchar_t, 1114111, 0>>' is deprecated` 
`'wstring_convert<std::codecvt_utf8<wchar_t, 1114111, 0>>' has been explicitly marked deprecated here`

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -3017,11 +3017,32 @@ namespace jwt {
 
 			static std::string to_lower_unicode(const std::string& str, const std::locale& loc) {
 #if __cplusplus > 201103L
-				std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> conv;
-				auto wide = conv.from_bytes(str);
+				std::mbstate_t state = std::mbstate_t();
+				const char* in_next = str.data();
+				const char* in_end = str.data() + str.size();
+				std::wstring wide;
+
+				while (in_next != in_end) {
+					wchar_t wc;
+					std::size_t result = std::mbrtowc(&wc, in_next, in_end - in_next, &state);
+					if (result == static_cast<std::size_t>(-1) || result == static_cast<std::size_t>(-2))
+						break; // conversion error
+					in_next += result;
+					wide.push_back(wc);
+				}
+
 				auto& f = std::use_facet<std::ctype<wchar_t>>(loc);
 				f.tolower(&wide[0], &wide[0] + wide.size());
-				return conv.to_bytes(wide);
+
+				std::string out;
+				for (wchar_t wc : wide) {
+					char mb[MB_CUR_MAX];
+					std::size_t n = std::wcrtomb(mb, wc, &state);
+					if (n != static_cast<std::size_t>(-1))
+						out.append(mb, n);
+				}
+
+				return out;
 #else
 				std::string result;
 				std::transform(str.begin(), str.end(), std::back_inserter(result),


include/jwt-cpp/jwt.h
```diff
@@ -3017,11 +3017,32 @@ namespace jwt {
...
+				std::mbstate_t state = std::mbstate_t();
+				const char* in_next = str.data();
+				const char* in_end = str.data() + str.size();
+				std::wstring wide;
+
+				while (in_next != in_end) {
+					wchar_t wc;
+					std::size_t result = std::mbrtowc(&wc, in_next, in_end - in_next, &state);
+					if (result == static_cast<std::size_t>(-1) || result == static_cast<std::size_t>(-2))
+						break; // conversion error
```
prince-chrismc, line 3029: Should his be an exception?

prince-chrismc:
For the linters there's a command you can run :) https://github.com/Thalhammer/jwt-cpp/actions/runs/4939213168/jobs/8838847109?pr=290 its just a copy paste from the logs

prince-chrismc:
Found this gem doing some research on the issue 🤣 

>The replacement is praying for C++23 to have proper Unicode support

> [3 yr. ago](https://www.reddit.com/r/cpp/comments/es9gt1/comment/ffcswsn/?utm_source=reddit&utm_medium=web2x&context=3)
!remindme 2023

***

https://isocpp.org/files/papers/D2139R3.html#3.21

Seems like there's no replacement but the tests are passing so 🤞

Thalhammer (approved):
Looks good, except a small nitpick and a question.

Given that the vast majority of data going through this function will be ascii, all three of `str`, `wide` and `out` will usually have the same size and it might be a good idea to call `reserve` on `wide` and `out` with the input size to prevent reallocations. With reserve worst case we overallocate a bit (on `wide`) or have to reallocate once (for `out`).

The other, bigger thing I am worried about is whether its really a good idea to replace the old/deprecated `wstring_convert` with the even older `mbrtowc` C api (which by the way is defined in `cwchar`, not in `cstring`. While I agree that we shouldn't use deprecated functions if it can be avoided, the `wstring_convert` code is just so much cleaner and I just can't see them removing it before there's an usable replacement.

EDIT: Oh and linting. you can just apply the patch command thats in the details of the failed ci run.

prince-chrismc:
> replace the old/deprecated wstring_convert with the even older mbrtowc C api

Would it make sense to drop the the C++11 ifdef? Or have a seperate >= C++17 section?

Thalhammer:
> > replace the old/deprecated wstring_convert with the even older mbrtowc C api
> 
> Would it make sense to drop the the C++11 ifdef? Or have a seperate >= C++17 section?

If we go with this solution we can certainly drop the ifdef because the c-style functions should be available in all C++ versions.

Thalhammer:
@laoshanxi Do you still need support for GCC 4.8 ?
I cancelled the check because the 18.04 ubuntu image has been [deprecated end of 2022](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/) and apparently has been removed now because it can't find a runner anymore. If you still need it, could you check if those changes work in your setup ?

laoshanxi:
> @laoshanxi Do you still need support for GCC 4.8 ? I cancelled the check because the 18.04 ubuntu image has been [deprecated end of 2022](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/) and apparently has been removed now because it can't find a runner anymore. If you still need it, could you check if those changes work in your setup ?

Let me have a try.

PR #292, opened 2023-05-16
Move gcc 4.8 tests to docker image

prince-chrismc:


.github/workflows/targets.yml
```diff
@@ -16,18 +16,25 @@ on:
...
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:18.04
     steps:
-      - run: sudo apt-get install g++-4.8
-      - uses: lukka/get-cmake@latest
+      - run: |
+          apt-get update
+          apt-get install -y g++-4.8
```
JonasBorchelt, line 25: I guess you know it by yourself, just add `wget`

Thalhammer:
Some of the tests failed because azure had a hickup, but the one we changed seems to be ok again.

prince-chrismc:
https://github.com/Thalhammer/jwt-cpp/actions/runs/5000111463/jobs/8957133536?pr=292

CMake Error at Source/kwsys/cmake_install.cmake:46 (file):
  file cannot create directory: /usr/local/doc/cmake-3.26/cmsys.  Maybe need
  administrative privileges.

prince-chrismc:
Ohh I just saw https://github.com/Thalhammer/jwt-cpp/pull/292/commits/eba0ef0d27480b5d3686a54fea9a5d01cce81c68 so that means we need to be smarter 🤔

prince-chrismc:
I am note sure why openssl 1.1.0 failed last time :/ https://github.com/Thalhammer/jwt-cpp/actions/runs/5006258717/jobs/8971288052#step:5:3138 

```
created directory `/tmp/bin'
install libcrypto.so.1.1 -> /tmp/lib/libcrypto.so.1.1
*** Installing runtime files
install libcrypto.so.1.1 -> /tmp/lib/libcrypto.so.1.1
link /tmp/lib/libcrypto.so -> /tmp/lib/libcrypto.so.1.1
chmod: cannot access '/tmp/lib/libcrypto.so.1.1.new': No such file or directory
make: *** [Makefile:387: install_runtime] Error 1
make: *** Waiting for unfinished jobs....
install libssl.so.1.1 -> /tmp/lib/libssl.so.1.1
link /tmp/lib/libssl.so -> /tmp/lib/libssl.so.1.1
```

It's working now 🤷

.github/workflows/cmake.yml
```diff
@@ -75,30 +75,14 @@ jobs:
   min-req:
     runs-on: ubuntu-20.04
     env:
-      RUN_ON: ubuntu-20.04
       CMAKE_URL: https://cmake.org/files/v3.14/cmake-3.14.7.tar.gz
       CMAKE_VERSION: 3.14.7
```
Thalhammer, line 79: Those are unused now, aren't they ?

.github/actions/install/cmake/action.yml
```diff
@@ -0,0 +1,39 @@
+name: Install CMake
+description: Download, Build and Cache CMake
+inputs:
+  version:
+    description: The desired CMake version to install
+    required: true
+  url:
```
Thalhammer, line 7: Is there a reason why we can't build the url from the version string ?
CMake's download urls seem to be fairly consistent and it would allow us to remove the url parameter or at least make it optional.

Thalhammer (changes requested):

.github/workflows/cmake.yml
prince-chrismc: You are absolutely right!

.github/actions/install/cmake/action.yml
prince-chrismc: It had the version twice in two different formats and I didn't take the time to parse and reformat it but it should be doable

Thalhammer: Ahh ok, don't bother, just wondered if we could merge them if they are always the same, but if they can differ it makes sense to keep both.

prince-chrismc: It's 1 version just written two different way :) 
The URL has both `3.14` and `3.14.7` so it needs work getting ride of the patch version. 

We shouldn't be compiling from sources often so didn't seem worth it

Issue #160, opened 2021-08-31
[Question] How were the JWKS and token generated for the jwks-verify.cpp example?


gerardoprada:
Can a JWK be generated from an RSA public key using this repo?

prince-chrismc:
> How were the JWKS and token generated for the jwks-verify.cpp example?

I am assuming the original author 
1. generate key/cert from the [command line with OpenSSL](https://stackoverflow.com/a/56479854/8480874)
2. base64 encoding
3. [extracted the values with openssl](https://stackoverflow.com/a/3117100/8480874)

> Can a JWK be generated from an RSA public key using this repo?

Sadly not yet, however if you are interested in contributing... I did some work on this and I can share my notes

gerardoprada:
> I am assuming the original author

Makes sense. Normally I generate my keys with OpenSSH//OpenSSL in terminal and if needed encode in PEM but hadn't worked with JWKS before. Was mostly curious as to whether they had been generated in code using this repo. Thanks for the answer.

> Sadly not yet, however if you are interested in contributing... I did some work on this and I can share my notes

Yeah feel free to share. Would be interesting to look at. I'd be happy to explore further.

prince-chrismc:
I have a test "pem to pem" which is for the RSA JWK scenario making sure I can remake a public key from the modulus and exponent (since `"n"` and `"e"` are required unlike `"x5c"` thats the workflow I had done)

The bulk should just be parsing/formatting the structure

- parsing
```cpp
   std::unique_ptr<BIGNUM, decltype(&BN_free)> base64DecodeBigNum(const std::string& base64bignum)
   {
      const auto decode = jwt::base::decode<jwt::alphabet::base64url>(jwt::base::pad<jwt::alphabet::base64url>(base64bignum));

      return jwt::helper::raw2bn(decode);
   }

   // https://gist.github.com/polesen/2855098
   // https://stackoverflow.com/a/10903704/8480874
   std::unique_ptr<EVP_PKEY, decltype(&EVP_PKEY_free)> RsaPubKeyFromModExp(const std::string& modulus_b64, const std::string& exp_b64)
   {
      auto n = base64DecodeBigNum(modulus_b64);
      auto e = base64DecodeBigNum(exp_b64);

      if (e && n)
      {
         EVP_PKEY* pRsaKey = EVP_PKEY_new();
         RSA* rsa = RSA_new();
         RSA_set0_key(rsa, n.release(), e.release(), nullptr);
[...]

PR #298, opened 2023-07-30
Add helper for making RSA key from exponent and modulus

### LINKED ISSUE #271: Support for JWT verification without x5c.
I have a use case where jwk doesn't have x5c but only `e` and `n` for RSA256. I noticed that it's not possible to only pass `e` and `n` to the library and get the jwt verified.

1. Does the library assume that we use OpenSSL methods to create RSA key with `e` and `n` and then user it for verification or am I missing something?
2. I did try that and I keep getting `error:04067072: rsa routines:r sa_ossl_public_decrypt: padding check failed`. I am far from an OpenSSL expert but I tried using all padding methods and none works. How do I know the padding method used in signature?

Thanks in advance.

zofer1:
While evaluating JWT-CPP I have found a case where we have the public key defined as modulus and exponent. I have added a wrapper function to allow this functionality and hide the openssl details for this.

I have also added adding a set of claims defined as json to a verifier to allow static configuration files in the application level.

For the case where we may apply external claims verification we only want to verify the signature only and skip the claims verification. For this I have split the verify functions accordingly while maintaining a BWC.

closes #271

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -23,6 +23,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <climits>
 #include <cmath>
 #include <cstring>
 #include <functional>
@@ -3042,7 +3043,7 @@ namespace jwt {
 				std::string out;
 				out.reserve(wide.size());
 				for (wchar_t wc : wide) {
-					char mb[MB_CUR_MAX];
+					char mb[MB_LEN_MAX];
 					std::size_t n = std::wcrtomb(mb, wc, &state);
 					if (n != static_cast<std::size_t>(-1)) out.append(mb, n);
 				}


prince-chrismc:
@zofer1 this looks really interesting, would you mind splitting the two features in 2 PRs? 

The exp+mod has been requested in the past #271and #160 and this looks similar to what I've suggested... however it is missing unit tests which I would live to see :) 

The second half, I am a little hesitant on

> and skip the claims verification

What's missing from the current verify that it's not suitable? 

You changes look minimal but I would rather not expose an API where devs can make easy mistakes that introduce security vulnerabilities

zofer1:
I can split it, not an issue and also add some unit test. It will take some time to do that.
I understand the security concern given that the entire specs around this JWT is very generic and basically says that security concerns must be addressed by implementation. However, if one have an implementation with heavy logic of claims processing this might not be a good fit for the hook method of verify_check_fn_t.
For this reason I thought it would be a good idea to either do just signature verification or with the basic exp+nbf.
For me this can be a "nice to have" thing.
The last addition I do think that can help is adding static verifiers from a json.

Thalhammer:
Hi,
thanks for your additions.

While I think the possibility to construct a key from components, as well as the split is a nice thing, I don't think the `with_claims` function fits with jwt-cpp. Loading a config feels like something that should be handled by the application.

zofer1:
> Hi, thanks for your additions.
> 
> While I think the possibility to construct a key from components, as well as the split is a nice thing, I don't think the `with_claims` function fits with jwt-cpp. Loading a config feels like something that should be handled by the application.

The inspiration for this is the parse_jwks function. While this is based on rfc7517#section-5 there is nothing about claim set.
Of course application can do this logic but relying on the heavy templating seems nicer from the inside of the package.

I do have additional wrapper in the application layer that is doing additional work. this work results in feeding a verifier with subsequent calls to with_claim which I suggest to replace with call to with_claims. This gives automatic handling for the special cases and saves the effort to handle this specifically in application code.

CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ endif()
 project(jwt-cpp)
 
 option(JWT_BUILD_EXAMPLES "Configure CMake to build examples (or not)" ON)
-option(JWT_BUILD_TESTS "Configure CMake to build tests (or not)" OFF)
+option(JWT_BUILD_TESTS "Configure CMake to build tests (or not)" ON)
```
prince-chrismc, line 19: The idea here is that consumer of the project can simply install the project which getting bother by the extra requirements of the tests which is why this is off by default

```suggestion
option(JWT_BUILD_TESTS "Configure CMake to build tests (or not)" OFF)
```

include/jwt-cpp/jwt.h
```diff
@@ -845,6 +847,74 @@ namespace jwt {
 			return pkey;
 		}
 
+		inline void freem(RSA*){}
```
prince-chrismc, line 850: left of test code? this seems out of place

tests/PubKeyTest.cpp
```diff
@@ -0,0 +1,25 @@
...
+)";
+
+std::string modulus = R"(AMKAd4lZuEVvukrZEfp7rcdX0Hr7tvrdBbuigXG74X8h0l8u8NJyTnU0-I1i40raUZDUAT2cDMBx9-YvttYHZybQ3v8XzvCF_THBZsqHZQVHKl_Aq7iMw7_QF39jo1zwRvuGqvtN1ypef5rgE5d9vvt9NVcNXV6BmDXqFkKi07B091kt7Tjn_nobszbmfq4_nqYWg95TAU6BAK67QvUfdSk0zemEgDiuPDcUwPAnzjBSuYrcXyKgefhPTkkE4nV8qi8qHgPscUyjKmH8b8qRHpNaLngIWPbuuzQgXZrmr8bX8r8Ke_qOknfjbHsMQIZkShXscNdyjmMw4QvvWjCXLiU)";
+std::string exponent = R"(AQAB)";
+
+TEST(PubKeyTest,RSAFromComponents){
+    auto pubkey = jwt::helper::create_public_key_from_rsa_components(modulus, exponent);
+
+    std::cout << pubkeypem << std::endl;
+    std::cout << pubkey << std::endl;
```
prince-chrismc, line 23: no need to debug code 

```suggestion
```

tests/CMakeLists.txt
```diff
@@ -55,7 +56,7 @@ endif()
 target_link_libraries(jwt-cpp-test PRIVATE jwt-cpp nlohmann_json::nlohmann_json
                                            $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:${CMAKE_DL_LIBS}>)
 
-gtest_add_tests(TARGET jwt-cpp-test)
+gtest_discover_tests(jwt-cpp-test)
```
prince-chrismc, line 59: 👏 https://cmake.org/cmake/help/latest/module/GoogleTest.html newer is better

and it's supported already https://github.com/Thalhammer/jwt-cpp/blob/6fed66d0d54911dd4a9eedd71abe745f7a3d9bd5/CMakeLists.txt#L1C37-L1C37

tests/CMakeLists.txt
```diff
@@ -34,7 +35,7 @@ add_executable(jwt-cpp-test ${TEST_SOURCES})
 
 # NOTE: Don't use space inside a generator expression here, because the function prematurely breaks the expression into
 # multiple lines. https://cmake.org/pipermail/cmake/2018-April/067422.html
-set(JWT_TESTER_GCC_FLAGS -Wall -Wextra -Wpedantic)
+set(JWT_TESTER_GCC_FLAGS -Wall -Wextra -Wpedantic -ggdb)
```
prince-chrismc, line 38: 🤔 https://stackoverflow.com/a/14811348/8480874

Since this is the tester, It makes sense but I am a fan of less is more

Thalhammer: Shouldn't cmake add flags to generate debug symbols by itself if you build in the debug configuration ?

zofer1:
Sorry I am new to working with GitHub so I thought I only have draft of the pull request and I knew there are some additional changes I need to make. I have created a new PR with relevant changes, I hope you find it better :-).
This PR includes just the first part we have agreed on, that is the create RSA pub key from components including unit test

include/jwt-cpp/jwt.h
```diff
@@ -3283,8 +3353,9 @@ namespace jwt {
 				return;
 			}
 			algs.at(algo)->verify(data, sig, ec);
+			
 			if (ec) return;
-
+
```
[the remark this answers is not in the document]
zofer1: Fixed that

include/jwt-cpp/jwt.h
zofer1: This was a temporary workaround to full the auto pointers

CMakeLists.txt
zofer1: Yes, I have reverted this when testing was completed

tests/CMakeLists.txt
prince-chrismc: It does `-g0` which is more portable, this is a compiler specific one that has extras

include/jwt-cpp/jwt.h
```diff
@@ -861,19 +860,19 @@ namespace jwt {
...
-						  static_cast<int>(decoded_modulus.size()), nullptr),
-				BN_free);
+			BIGNUM* n = BN_bin2bn(reinterpret_cast<const unsigned char*>(decoded_modulus.data()),
+						  static_cast<int>(decoded_modulus.size()), nullptr);
 
-			std::unique_ptr<BIGNUM, decltype(&BN_free)> e(
-				BN_bin2bn(reinterpret_cast<const unsigned char*>(decoded_exponent.data()),
-						  static_cast<int>(decoded_exponent.size()), nullptr),
-				BN_free);
+			BIGNUM* e = BN_bin2bn(reinterpret_cast<const unsigned char*>(decoded_exponent.data()),
```
prince-chrismc, line 866: Why did you remove the unique ptr? According to https://www.openssl.org/docs/man3.1/man3/BN_bin2bn.html it needs to be freed 🤔

include/jwt-cpp/jwt.h
```diff
@@ -861,19 +860,19 @@ namespace jwt {
...
+						  static_cast<int>(decoded_exponent.size()), nullptr);
 
-			std::unique_ptr<RSA, decltype(&freem)> rsa(RSA_new(), freem);
-			if (RSA_set0_key(rsa.get(), n.get(), e.get(), nullptr) != 1){
+			std::unique_ptr<RSA, decltype(&RSA_free)> rsa(RSA_new(), RSA_free);
+
+			//This is unlikely to fail, and after this call RSA_free will also free the n and e big numbers
+			if (RSA_set0_key(rsa.get(), n, e, nullptr) != 1){
 				ec = error::rsa_error::set_rsa_failed;
+				BN_free(e);
```
prince-chrismc, line 874: Ahhhh

> The n, e and d parameter values can be set by calling RSA_set0_key() and passing the new values for n, e and d as parameters to the function. The values n and e must be non-NULL the first time this function is called on a given RSA object. The value d may be NULL. On subsequent calls any of these values may be NULL which means the corresponding RSA field is left untouched. Calling this function transfers the memory management of the values to the RSA object, and therefore the values that have been passed in should not be freed by the caller after this function has been called.

Now I understand the double free comment

prince-chrismc (commented):
I'd suggest keeping the unique ptr and perhaps 

```cpp
RSA_set0_key(rsa.get(), n.release(), e.release(), nullptr)
```

I noticed that's the original suggestion https://github.com/Thalhammer/jwt-cpp/pull/298#pullrequestreview-1560197150

prince-chrismc:
I just noticed that the 1.0.2 CI failed and looking into it, I noticed this function was not invented then https://www.google.com/search?sitesearch=www.openssl.org&q=RSA_set0_key

so it will need some `ifdef` logic to disable this function

dr0pdb:
Just FYI since it might be of interest, I ended up implementing this for RSA and EC keys. Link for the code is in this comment - https://github.com/Thalhammer/jwt-cpp/issues/271#issuecomment-1673084618. So may be it can used as a reference for adding the support for EC keys as well.

zofer1:
> Just FYI since it might be of interest, I ended up implementing this for RSA and EC keys. Link for the code is in this comment - [#271 (comment)](https://github.com/Thalhammer/jwt-cpp/issues/271#issuecomment-1673084618). So may be it can used as a reference for adding the support for EC keys as well.

Yes I wanted to get to that sometime. For me there is a learning curve in both cryptography and open source contribution so I planned to do this after the first PR is approved. Now that I see it takes time, I will probably get to that by the time it is approved and add it here.

prince-chrismc:
derp I broke something

PR #304, opened 2023-08-12
a little change to make it more easy to use

Derrity:
then we can use 
```cpp
auto decoded = jwt::decode(token)
```
and
```cpp
bool verifile = jwt::verify()
```


Derrity:
```cpp
const bool verify(const decoded_jwt<json_traits>& jwt) const {
            try {
                std::error_code ec;
                verify(jwt, ec);
                error::throw_if_error(ec);
                return true;
            }catch(const std::exception&){
                return false;
            }
		}
```

prince-chrismc:
A bool is not capable of transmitting all the various error states.


This would be a welcomed addition if you followed the existing `std::error_code` mechanism the library implemented

PR #306, opened 2023-08-17
Add using namespace jwt to unit tests

sjanel:
All is in the title. No other change in this PR.

### DIFF
--- a/tests/BaseTest.cpp
+++ b/tests/BaseTest.cpp
@@ -1,110 +1,112 @@
 #include "jwt-cpp/base.h"
 #include <gtest/gtest.h>
 
+using namespace jwt;
+
 TEST(BaseTest, Base64Index) {
-	ASSERT_EQ(0, jwt::alphabet::index(jwt::alphabet::base64::data(), 'A'));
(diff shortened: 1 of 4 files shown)


Thalhammer:
Ok, but what's the benefit of doing this ?
The tests aren't supposed to be included in other programs, so name clashes shouldn't be an issue.

sjanel:
Benefits are readability, and removal of all prefixes of symbols accessed in jwt (removal of `jwt::`).

prince-chrismc:
> Benefits are readability, and removal of all prefixes of symbols accessed in jwt (removal of `jwt::`).

Why not just stick a `using namespace` at the top?

This seems very stylized and adds a lot of diff, why are you looking at the symbols of the test code? You are never going to link against it.

sjanel:
I don't have a strong opinion on this subject. Putting a top using namespace is fine for me.

prince-chrismc:
Thanks a ton for this suggestion, since we seem to be agreed this is not a "functional" improvement and we have split or indifferent opinions... I am going to side of the side of caution and mark this as closed.

I really appreciate you brining in new ideas, please keep them coming! 🚀

PR #307, opened 2023-08-21
 Add helper for making RSA key from exponent and modulus

prince-chrismc:
Follow up of #298 since GitHub closed the PR when I pushed and will not let me re-open 😞 

I fixed the wolfSSL test not passing, I put together some code for the openssl 3.0 support

//cc @zofer1 who originally submitted this

***
While evaluating JWT-CPP I have found a case where we have the public key defined as modulus and exponent. I have added a wrapper function to allow this functionality and hide the openssl details for this.

I have also added adding a set of claims defined as json to a verifier to allow static configuration files in the application level.

For the case where we may apply external claims verification we only want to verify the signature only and skip the claims verification. For this I have split the verify functions accordingly while maintaining a BWC.

closes https://github.com/Thalhammer/jwt-cpp/issues/271


zofer1:
I see you made a lot of changes. OpenSSL did deprecate the functions in v3 but it is still usable. I will test the openssl 3 changes although I wanted to do this at a later stage after the original PR is approved and with the additions for elliptic curve.
Please advise if further tasks are pending so we make it this time before the PR is closed

prince-chrismc:
We made a decent effort to remove all the deprecated function usage with OpenSSL 3.0 and it would be ideal to keep it that way. The test code you submitted passes but I do agree it needs to be tested more.

I would love to fixup some of the memory management, there's a current a leak when it makes the key. it would be easier to merge this before trying to add more.

I think it's getting a little bit complicated and the EC curve will need just a much fiddling to maintain support for the 5 (or 6) combination of SSL libraries we support

include/jwt-cpp/jwt.h
```diff
@@ -846,6 +895,164 @@ namespace jwt {
...
+				ec = error::rsa_error::create_mem_bio_failed;
+				return {};
+			}
+
+			auto write_pem_to_bio =
+#if defined(JWT_OPENSSL_3_0)
+				// https://www.openssl.org/docs/man3.1/man3/PEM_write_bio_RSA_PUBKEY.html
+				&PEM_write_bio_PUBKEY;
+#else
+				&PEM_write_bio_RSA_PUBKEY
```
prince-chrismc, line 990:
```suggestion
				&PEM_write_bio_RSA_PUBKEY;
```

prince-chrismc:
I am pretty happy with this there's some missing negative case tests https://coveralls.io/builds/62156107/source?filename=include%2Fjwt-cpp%2Fjwt.h#L953 especially for the openssl error and I try to find sometime this week to work on those

include/jwt-cpp/jwt.h
```diff
@@ -846,6 +895,164 @@ namespace jwt {
...
+		* \param modulus	string containing base64 encoded modulus
+		* \param exponent	string containing base64 encoded exponent
+		* \param decode 				The function to decode the certs)
+		* \param ec			error_code for error_detection (gets cleared if no error occur
+		* \return public key in PEM format
+		*/
+		template<typename Decode>
+		std::string create_public_key_from_rsa_components(const std::string& modulus, const std::string& exponent,
+														  Decode decode, std::error_code& ec) {
+			auto decoded_modulus = decode(modulus);
```
Thalhammer, line 910: Would it make sense to pull out the decode template and put that burden on the caller ?
It's just a small nitpick, but it will reduce the binary size of apps that have lots of different decode functions.

Thalhammer (approved):
Except for the tiny nitpick it looks fine.

include/jwt-cpp/jwt.h
zofer1: In my mind, we will get an external definition of public key in the jwks and jwk format and creating the public key is already a helper function. As such, I think using another helper (twice) to a helper function is a little cumbersome.

I am not concerned about compiled code size. These days it can be a concern in embedded systems that are tiny but nothing is tiny anymore

prince-chrismc: I suspect if they are at the level of binary size being a concern they will have already implemented thier own highly optimized base64 decoder and if they call that one - they do not pay the price if the lambda.

This was originally a way to keep the API stable while also keeping the base.h header as optional 

I would definitely be will to change that for the next release if we want to simplify or separate some of the code to new headers

prince-chrismc:
Finally found the memory leak 🚀 

More improvements to actions + fixed an example 😫 worth it

prince-chrismc:
Finally green ✅

Issue #305, opened 2023-08-14
Create a test for the undefined sanitizer


prince-chrismc:
This is something we are missing and can help avoid bugs

https://github.com/Thalhammer/jwt-cpp/issues/300#issue-1846108795

PR #315, opened 2023-11-28
Add a test for ubsan with C++20

prince-chrismc:
closes https://github.com/Thalhammer/jwt-cpp/issues/305


prince-chrismc:
I was considering https://github.com/ammaraskar/gcc-problem-matcher but I am not sure about all the different outputs so I've opted for the sake of CI to just abort on the first error

PR #317, opened 2023-12-09
Implemented JsonCPP Traits

cjserio:
My project already makes heavy use of json-cpp so I put together this implementation of json-cpp traits for the jwt-cpp library since it seems like others have been asking for it as well.


prince-chrismc:
Thank you so much for this great addition! This bit requires a bunch of CI to ensure consistency so I took the liberty of adding (most of it 🤞) there are a hand full of linter checks that are failing and I suspect there will be some compilation failures... especially in the test which I am not familiar with the API of this library.

Any help fixing it up will be appreciated!

I took some liberties with naming just to make it `author-library` without making it too long, open to suggestions but I know that name will eventually have other very similar projects :)

prince-chrismc:
API does not match exactly but seems very close https://github.com/Thalhammer/jwt-cpp/actions/runs/7153838549/job/19480726737?pr=317#step:11:1786

cjserio:
You're welcome! It's a really popular library so I know it'll be helpful to people. I can take a look at the API issues this week if you'd like.

prince-chrismc:
I think a future task will be to more the "linters" from nodejs to CMake so thier are more friendly to run locally because it's kinda a pain

Issue #186, opened 2021-11-19
Improvement to key loading helpers


prince-chrismc:
We should deprecate/remove the older helpers and rename them to have RSA in the name.

_Originally posted by @prince-chrismc in https://github.com/Thalhammer/jwt-cpp/issues/185#issuecomment-973717147_

prince-chrismc:
> I also switched from [`pem_bio_read`]. This change made it obvious that one could use the same code for reading RSA and EC keys. The functions only differ in type of errors they report, e.g., error::ecdsa_error::create_mem_bio_failed vs. error::rsa_error::create_mem_bio_failed.

_Orignally posted in https://github.com/Thalhammer/jwt-cpp/pull/185#issue-1057747089_

prince-chrismc:
The newer EC family of helpers are also missing test code which was present with the old ones `load_public_ec_key_from_string` for example

prince-chrismc:
The implementations are nearly the same they just need different error codes but otherwise are interchangable

```diff
diff --git a/include/jwt-cpp/jwt.h b/include/jwt-cpp/jwt.h
index 000fb0b..1235585 100644
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -778,7 +778,7 @@ namespace jwt {
                        ec.clear();
                        auto pubkey_bio = make_mem_buf_bio();
                        if (!pubkey_bio) {
-                               ec = error::ecdsa_error::create_mem_bio_failed;
+                               ec = error::rsa_error::create_mem_bio_failed;
                                return {};
                        }
                        if (key.substr(0, 27) == "-----BEGIN CERTIFICATE-----") {
@@ -786,13 +786,13 @@ namespace jwt {
                                if (ec) return {};
                                const int len = static_cast<int>(epkey.size());
                                if (BIO_write(pubkey_bio.get(), epkey.data(), len) != len) {
-                                       ec = error::ecdsa_error::load_key_bio_write;
+                                       ec = error::rsa_error::load_key_bio_write;
                                        return {};
                                }
                        } else {
                                const int len = static_cast<int>(key.size());
                                if (BIO_write(pubkey_bio.get(), key.data(), len) != len) {
-                                       ec = error::ecdsa_error::load_key_bio_write;
+                                       ec = error::rsa_error::load_key_bio_write;
                                        return {};
                                }
                        }
@@ -800,7 +800,7 @@ namespace jwt {
[...]

PR #318, opened 2023-12-10
Remove duplicated key handlers

prince-chrismc:
LoadPublicKeyFromStringReferenceWithEcCert

Trying to test https://github.com/Thalhammer/jwt-cpp/issues/186 to see if it's easy


prince-chrismc:
in https://github.com/Thalhammer/jwt-cpp/pull/318/commits/5fd1ad9ad585e47d255dea3d8cd16d59d721b65b I noticed EdDSA is using a mixture of RSA and ECDSA error codes 😦

Thalhammer (approved):
I like the general idea of cleaning up the error handling, given that rsa_error is kind of wrong if you load an ecdsa key.

However I am not sure if I like the implementation. You are essentially duplicating every key error into ecdsa and rsa and the user has to care about that, even though the underlying error is basically the same. With all the upcoming/planned work towards jwks, key related error cases and codes are only going to get more which makes me feel/wonder if a separate error category for key handling might be a better fit. Especially given a key class would very likely be capable of holding all kinds of keys (rsa, ecdsa, maybe even symetric) and I really don't care which one was inside when I want to convert it to pem.

I am still approving this, since the code on its own looks good and it definitly is better than the current solution. I am just not sure if its the best solution.

prince-chrismc:
I feel the same way, I was considering fixing the eddsa keys being both RSA and EC -- the big think for 0.8.0 will likely be improving the error handling starting with #252 I am happy I did this because it gave me a better idea of what we have and need to possibly change :+1:

I was thinking of a proper `cryoptographic_error` to cover all of them similar to have there `signature_verification` and co.

Issue #319, opened 2023-12-15
Guide to Setting Up jwt-cpp?


hkRho:
### What's your question?

Guide to Setting Up jwt-cpp?

### Additional Context

Hi, before anything, thank you for creating and supporting this repo!

I was a bit confused in using this library--should I just include the entire folder in the desired project folder that I wish to use jwt-cpp? Do I also have to install the dependencies that I'd like to use (like OpenSSL) separately on my own, and/or build the repo with cmake? 

I'm trying to use the functionalities in Visual Studio, but the #include is not recognizing jwt-cpp's dependencies like Openssl and picojson so was a bit lost on the setup steps that I may be missing.
![image](https://github.com/Thalhammer/jwt-cpp/assets/42980938/bf5c8865-6bbb-4f87-b31e-68dae401a4f4)
I've tried including the jwt-cpp repo under my project directory and tried building it using cmake (the build seemed to work except this part `-- Could NOT find jsoncons (missing: jsoncons_DIR) -- Could NOT find nlohmann_json (missing: nlohmann_json_DIR)` 

I really appreciate taking a look at this basic question and would love any guidance on how I can properly set jwt-cpp up and use it!

Thalhammer:
Hi,

Jwt-cpp is extremly flexible when it comes to your project setup, all you need to make sure is that you have the contents in the `include` of this repo available in your include path and at least one of the dependencies is available.

You can, but don't have to use make for this. Assuming your project already uses cmake it just makes it extremly easy because you can use fetchcontent or a cmake package manager like cpm or hunter to pull in jwt-cpp.

However since you seem to be using visual studio without cmake, I would recommend using the [official nuget packages](https://www.nuget.org/packages/jwt-cpp) for both jwt-cpp and openssl since they will setup all the include directories and libraries as needed.

Picojson is the only dependency thats shipped with jwt-cpp, but its mostly a placeholder for those that do not use another json library already, nowadays since the developer of it basically abandonded it years ago. Assuming you have set up your include paths correctly it should get recognized, since its part of the same directory.

hkRho:
Hi @Thalhammer,

Thank you for a quick reply!

Just a quick clarification on 

> "make sure is that you have the contents in the include of this repo available in your include path"

so if I had `#include <openssl/ssl.h>` in `jwt.h`, I'd have to make sure that the ssl.h exists in `jwt-cpp\include`?
[...]

PR #321, opened 2023-12-25
Build examples only if master project by default

bugdea1er:
Hi there!

I changed the default value of JWT_BUILD_EXAMPLES to true iff jwt is the main project.

We are using jwt-cpp as a submodule, and although we have manually disabled this option, it is strange to build examples in a case with a submodule where this is not explicitly set.

What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,14 @@ endif()
 
 project(jwt-cpp)
 
-option(JWT_BUILD_EXAMPLES "Configure CMake to build examples (or not)" ON)
+# Determine if this is built as a subproject (using add_subdirectory)
+if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  set(JWT_MASTER_PROJECT ON)
+else()
+  set(JWT_MASTER_PROJECT OFF)
+endif()
+
+option(JWT_BUILD_EXAMPLES "Configure CMake to build examples (or not)" ${JWT_MASTER_PROJECT})
 option(JWT_BUILD_TESTS "Configure CMake to build tests (or not)" OFF)
 option(JWT_ENABLE_COVERAGE "Enable code coverage testing" OFF)
 option(JWT_ENABLE_FUZZING "Enable fuzz testing" OFF)


CMakeLists.txt
```diff
@@ -15,7 +15,14 @@ endif()
 
 project(jwt-cpp)
 
-option(JWT_BUILD_EXAMPLES "Configure CMake to build examples (or not)" ON)
+# Determine if this is built as a subproject (using add_subdirectory)
+if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
```
prince-chrismc, line 19: There's a modern way for this https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html

Not sure when it was added

prince-chrismc:
Personally not inclined to this, as you mentioned it was easy for you to set this on your end. My reasoning is users like #319 which might not know so it best to have examples compiled by default to help find problems.

Maybe if it was a single line or was more consistent in the ecosystem I would be swayed. This doesn't help for package managers either so it's a really slim user.

Open to my mind being changed!

CMakeLists.txt
bugdea1er: I don't think it's worth bumping the version from 3.14 for this feature alone

bugdea1er:
Honestly, I don't think a new user will need to compile the examples to understand how to use this library. Source codes usually help with this.

Maybe at least add a warning about this in the README? I mean instructions on how to disable example compilation?

This won't help people who know about it (like our team), it can only help new users.

prince-chrismc:
> Honestly, I don't think a new user will need to compile the examples to understand how to use this library. 

Ahh, I meant specific that it can build, because without anything to compile (being a header only project) its hard to know if the setup was correct. By having examples enabled by default, comments like this..

>> I've tried including the jwt-cpp repo under my project directory and tried building it using cmake (the build seemed to work 

..have a good amount, of information. Now that's not to say a CMake novice might be using the wrong words. My hope is that by making new users build them we know the system in properly configured (which is very often a pain point).

Hopefully that clarifies my last remarks 😄

prince-chrismc:
Sure, but using `add_subdirectory` and `fetchcontent` are not "officially" supported and there are no tests for them. My hesitation is looking at supporting these the [CMake CI](https://github.com/Thalhammer/jwt-cpp/actions/runs/7331412479) will just grow unreasonably from the POV of the maintainer (aka me). Now https://github.com/Thalhammer/jwt-cpp/commit/ca110ade5bf70920175bd6679c39e2c0f79239d7#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a I did change tests so I should be more open to examples. There are very few drawbacks to this suggestion.
GTest also recommends override settings https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/tests/CMakeLists.txt#L21-L22 so this is an option.

I do think improving the documentation in this area is super important and I've included your suggestion as I've been working on this.

image

However looking at my copy of Professional CMake 16th edition, I think I've talked myself in this is a good change, I am going to poke some friends and colleagues to get more opinions.

PR #322, opened 2023-12-27
Add set_expires_in and set_issued_now methods

bugdea1er:
Hi there!

I added `set_expires_in` and `set_issued_now` method for the jwt builder.

- `set_issued_now` sets a "iat" claim by calling system_clock::now() by itself;
- `set_expires_in` sets a "exp" claim by adding the given duration to ~~the "iat" claim~~ system_clock::now()

Common usage of the jwt builder is:
- create a builder
- set the "iat" claim at the "now" moment
- set the token expiration timeout

This should make it a little bit more convenient.

Update: builder now has the Clock template parameter just like the verifier

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -2744,6 +2744,19 @@ namespace jwt {
 		 */
 		builder& set_expires_at(const date& d) { return set_payload_claim("exp", basic_claim<json_traits>(d)); }
 		/**
+		 * Set expires at claim to @p d after the issued at claim
+		 * MUST be called after set_issued_at or set_issued_now
+		 * \param d token expiration timeout
+		 * \return *this to allow for method chaining
+         * \throw std::runtime_error If issued at claim was not present
+         * \throw std::bad_cast Claim was present but not a date (Should not happen in a valid token)
+		 */
+		template<class Rep>
+		builder& set_expires_in(const std::chrono::duration<Rep>& d) {
+			date iat = payload_claims["iat"].as_date();
+			return set_payload_claim("exp", basic_claim<json_traits>(iat + d));
+		}
+		/**
 		 * Set not before claim
 		 * \param d First valid time
 		 * \return *this to allow for method chaining
@@ -2755,6 +2768,11 @@ namespace jwt {
 		 * \return *this to allow for method chaining
 		 */
 		builder& set_issued_at(const date& d) { return set_payload_claim("iat", basic_claim<json_traits>(d)); }
+		/**
+		 * Set issued at claim to current time, as determined by std::chrono::system_clock::now()
+		 * \return *this to allow for method chaining
+		 */
+		builder& set_issued_now() { return set_issued_at(std::chrono::system_clock::now()); }
 		/**
 		 * Set id claim
 		 * \param str ID to set
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -2622,13 +2622,19 @@ namespace jwt {
 	 * Builder class to build and sign a new token
 	 * Use jwt::create() to get an instance of this class.
 	 */
-	template<typename json_traits>
+	template<typename Clock, typename json_traits>
 	class builder {
 		typename json_traits::object_type header_claims;
 		typename json_traits::object_type payload_claims;
 
+		/// Instance of clock type
+		Clock clock;
 	public:
-		builder() = default;
+		/**
+		 * Constructor for building a new builder instance
+		 * \param c Clock instance
+		 */
+		explicit builder(Clock c) : clock(c) {}
 		/**
 		 * Set a header claim.
 		 * \param id Name of the claim
@@ -2772,7 +2778,7 @@ namespace jwt {
 		 * Set issued at claim to current time, as determined by std::chrono::system_clock::now()
 		 * \return *this to allow for method chaining
 		 */
-		builder& set_issued_now() { return set_issued_at(std::chrono::system_clock::now()); }
+		builder& set_issued_now() { return set_issued_at(clock.now()); }
 		/**
 		 * Set id claim
 		 * \param str ID to set
@@ -3608,8 +3614,8 @@ namespace jwt {
 	 * Return a builder instance to create a new token
 	 */
 	template<typename json_traits>
-	builder<json_traits> create() {
-		return builder<json_traits>();
+	builder<default_clock, json_traits> create(default_clock c = {}) {
+		return builder<default_clock, json_traits>(c);
 	}
 
 	/**
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -3590,6 +3590,16 @@ namespace jwt {
 	}
 
 	/**
+	 * Create a builder using the given clock
+	 * \param c Clock instance to use
+	 * \return builder instance
+	 */
+	template<typename Clock, typename json_traits>
+	builder<Clock, json_traits> create(Clock c) {
+		return builder<Clock, json_traits>(c);
+	}
+
+	/**
 	 * Default clock class using std::chrono::system_clock as a backend.
 	 */
 	struct default_clock {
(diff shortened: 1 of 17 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -2755,6 +2768,11 @@ namespace jwt {
 		 * \return *this to allow for method chaining
 		 */
 		builder& set_issued_at(const date& d) { return set_payload_claim("iat", basic_claim<json_traits>(d)); }
+		/**
+		 * Set issued at claim to current time, as determined by std::chrono::system_clock::now()
+		 * \return *this to allow for method chaining
+		 */
+		builder& set_issued_now() { return set_issued_at(std::chrono::system_clock::now()); }
```
prince-chrismc, line 2775: For consistency, this should match the `verifier` and have a templated clock 

https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/include/jwt-cpp/jwt.h#L3249-L3250

include/jwt-cpp/jwt.h
```diff
@@ -2743,6 +2743,19 @@ namespace jwt {
 		 * \return *this to allow for method chaining
 		 */
 		builder& set_expires_at(const date& d) { return set_payload_claim("exp", basic_claim<json_traits>(d)); }
+		/**
+		 * Set expires at claim to @p d after the issued at claim
+		 * MUST be called after set_issued_at or set_issued_now
```
prince-chrismc, line 2748: This sounds error prone, I would make this relative to `now()` using a templated clock

prince-chrismc (commented):
I think these are good QOL improvements. Just all of the examples

```diff
	auto token = jwt::create()
		.set_issuer("auth0")
		.set_type("JWT")
		.set_id("rsa-create-example")
-		.set_issued_at(std::chrono::system_clock::now())
-		.set_expires_at(std::chrono::system_clock::now() + std::chrono::seconds{36000})
+		.set_issued_now()
+		.set_expires_in(std::chrono::seconds{36000})
		.set_payload_claim("sample", jwt::claim(std::string{"test"}))
		.sign(jwt::algorithm::rs256("", rsa_priv_key, "", ""));
``` 

I would love to bring the `verifier` and `builder` closer in API design by making them use the `default_clock` and this way under the hood both of them can have a `clock::now()` to help them.

I think it makes sense the build could do the same as https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/include/jwt-cpp/jwt.h#L3478

WDYT?

include/jwt-cpp/jwt.h
bugdea1er: If the "iat" claim was not set, should this method also set it to the "now" moment?

include/jwt-cpp/jwt.h
bugdea1er: This - you mean method or the `builder` class? If the clock is templated, shouldn't both `set_issued_now` and `set_expire_in` has the same clock source?

prince-chrismc: The class, this way it's constructed with the same clock :) the verifier does that already

include/jwt-cpp/jwt.h
prince-chrismc: 🤔 Regardless of if it is present it should be relative to the `clock.now()`. Having a single behavior would be the best approach

include/jwt-cpp/jwt.h
bugdea1er: Added a type parameter Clock to the builder

Doesn't compile at the moment tho, have to fix json traits now

include/jwt-cpp/jwt.h
bugdea1er: Changed this to
```cpp
template<class Rep>
builder& set_expires_in(const std::chrono::duration<Rep>& d) {
	return set_payload_claim("exp", basic_claim<json_traits>(clock.now() + d));
}
```

What type should I use for the duration parameter? Do we consider Clock::now() return such a type that a duration could be added to?

tests/traits/BoostJsonTest.cpp
```diff
@@ -83,6 +83,19 @@ TEST(BoostJsonTest, VerifyTokenExpirationValid) {
 	verify.verify(decoded_token);
 }
 
+TEST(BoostJsonTest, VerifyTokenExpirationInValid) {
```
bugdea1er, line 86: This needs a better name

include/jwt-cpp/jwt.h
bugdea1er: Done

bugdea1er:
Should I update examples or README with the new methods?

include/jwt-cpp/jwt.h
```diff
@@ -3590,8 +3610,8 @@ namespace jwt {
 	 * Return a builder instance to create a new token
 	 */
 	template<typename json_traits>
-	builder<json_traits> create() {
-		return builder<json_traits>();
+	builder<default_clock, json_traits> create(default_clock c = {}) {
```
prince-chrismc, line 3613: there also should be a create method that allows users to specify both types 

https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/include/jwt-cpp/jwt.h#L3765

include/jwt-cpp/traits/boost-json/defaults.h
```diff
@@ -25,9 +25,12 @@ namespace jwt {
...
 
 	/**
-	 * Return a builder instance to create a new token
+	 * Create a builder using the default clock
+	 * \return builder instance to create a new token
 	 */
-	inline builder<traits::boost_json> create() { return builder<traits::boost_json>(); }
+	inline builder<default_clock, traits::boost_json> create() {
+		return builder<default_clock, traits::boost_json>(default_clock{});
+	}
```
prince-chrismc, line 33: This also needs to be in the template 
https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/include/jwt-cpp/traits/defaults.h.mustache#L32

tests/traits/BoostJsonTest.cpp
prince-chrismc: We usually have an "expired" version of this test but with this API, I do not think that makes sense 

https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/tests/traits/TraitsTest.cpp.mustache#L92C15-L92C15

prince-chrismc (commented):
> Should I update examples or README with the new methods?

I think that would be excellent. Going over the code in more details, I think this new API is less error prone (prevents creating tokens that are already expired) and I'd love to see this being used going forward

include/jwt-cpp/jwt.h
prince-chrismc: Hmmm 🤔 currently it's done with `size_t` and assumes seconds

https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/include/jwt-cpp/jwt.h#L3331

I think this is totally fine, the clock has no concept so I think it's adequate since we already assume substraction https://github.com/Thalhammer/jwt-cpp/blob/953aab6fe47be9e92fcbf5fb8418dee0d682e7ca/include/jwt-cpp/jwt.h#L3154

include/jwt-cpp/jwt.h
bugdea1er: Done

include/jwt-cpp/jwt.h
```diff
@@ -2622,13 +2622,19 @@ namespace jwt {
 	 * Builder class to build and sign a new token
 	 * Use jwt::create() to get an instance of this class.
 	 */
-	template<typename json_traits>
+	template<typename Clock, typename json_traits>
```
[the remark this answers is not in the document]
bugdea1er: I can't imagine where it can be useful

include/jwt-cpp/traits/boost-json/defaults.h
bugdea1er: Done

prince-chrismc (approved):
Thanks for this awesome improvement!

PR #329, opened 2024-02-05
Remove unused header

ErwanLegrand:
The title tells it all. The codecvt header is no longer used in jwt-cpp therefore there is no reason to keep including it.


Thalhammer (approved):
Looks good to me. codecvt is deprecated in C++17 anyway and we don't use it anymore.

PR #334, opened 2024-03-17
support json string and picojson value type to set payload.

Unbreak4ble:
This will allow programmers to quickly set payload by using an existing json string.

### DIFF
--- a/example/set-payload.cpp
+++ b/example/set-payload.cpp
@@ -1,0 +1,27 @@
+/// @file set-payload.cpp
+#include <iostream>
+#include <jwt-cpp/jwt.h>
+
+int main() {
+	std::string json = R"(
+		{
+			"exp": 1711706773,
+			"iat": 1710706773,
+			"set": {
+				"complex": "json",
+				"payload": "using",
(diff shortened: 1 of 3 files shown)


prince-chrismc:
Could you please add some tests and an example? I would love to see how you are expecting this be be used :) 

For the example how would you set the `iat` and `exp` claims.

prince-chrismc:
My concern is this extra helper adds more complexity and could be mis-used.

How do you plan on changing those values such that they would be correct? The timestamps should always be different. Another's design questions, how do you prevent someone from destroying values they set?

What about the current API is not able to supporting this? You could roughly

```cpp
const auto json_object = traits::parse(your_str);
for(auto [key, value] : json_object) {
   jwt_builder.set_payload_claim(key, value);
}
```

Why is this "quickly"? parsing JSON is terribly slow.

Unbreak4ble:
The ideia is to parse a json string into the payload by using a single line. And i used the incorrect term "quickly" to express how big lines can be done in a single line.

I had not thought about using interactive loops. Thank you!

PR #342, opened 2024-04-18
Find nlohmann_json test dependency only when required

justend29:
nlohmann::json is used in the [examples](https://github.com/Thalhammer/jwt-cpp/blob/master/example/CMakeLists.txt#L33) and [tests](https://github.com/Thalhammer/jwt-cpp/blob/master/tests/CMakeLists.txt#L74) of jwt-cpp. However, it was always found, making it a transitive dependency for all users of jwt-cpp - even those not building examples or tests. These changes simply check if nlohman_json is required as a host dependency before searching for it.


prince-chrismc (approved):

Thalhammer (approved):
If @justend29 wants to extend it with `cmake_dependent_option` I am open to it, otherwise its good to merge as is.

justend29:
@prince-chrismc @Thalhammer 

Yes - I believe falling back to downloading and building the package when it's not found is not a convenience, it's a mistake. It's far too easy to mistype paths and then get the surprise of downloading a package you already have. Explicit selection of which to use is worthwhile.

A dependent option, maybe `JWT_EXTERNAL_TEST_NLOHMANN_JSON`, would offer this control. Additionally, CMake has built-in integration between `find_package` and `fetch_content` offered in CMake 3.24+ allowing standard CMake variables control how the dependencies are acquired, and dependency providers can be used to override it. Nevertheless, this project only mandates CMake 3.14+, and CMake 3.24's integrations would only complement JWT's variable for users who use compatible versions.

Consider it done.

Also, really great library. Thank you for the contributions.

prince-chrismc:
- none of these options are used by the package managers so 👍 
- the cmake installed config does not check for nlohmann
- the cmake tests use the internal find_package module for picojson

LGTM, thanks for the contribution. 

I think there are missing tests for and/or feature for pickup the different JSON libraries from out find config but thats also not how it's supported? it just [adds all of them](https://github.com/Thalhammer/jwt-cpp/blob/084c5fbbac09e3e43cdd9c5e5500a0c48278359c/CMakeLists.txt#L38C5-L38C28) and lets the user figure the rest out 🤷

PR #346, opened 2024-04-21
Default constructor for jwks object

justend29:
I appreciate the design to always uphold the invariants of every type. 
For `jwt::jwks`, which is an object holding a collection of jwks, constructing an empty collection of jwks is valid and maintains the invariants of the type.
The motivation for this is encapsulating `jwt::jwks` in a caching class, where JWKs are often cached with periodic refreshing. There aren't any keys until fetched from issuer. I found myself doing `jwt::jwks{"{\"keys\":[]}"}` since there wasn't a default constructor, which feels silly.


justend29:
whoever created the CI job to not only indicate exactly where the lining error is, but also generate a patch file and sample commands to apply it deserves a gold star.

prince-chrismc:
> also generate a patch file and sample commands to apply it deserves a gold star.

⭐ appreciate it 🙏

justend29:
@prince-chrismc I shouldn't have made the PR w/o a test. It's in.

prince-chrismc (approved):

PR #348, opened 2024-05-12
Add support for CMake Presets

prince-chrismc:
The intention is to move some of the setup and configuration variables from the CI scripts to the build configurations to make it easier to run them locally. This is most evident when it comes to the sanitizers and fuzzers but also for the examples. 

The second benefit is also from #344 is to make the cross platform tests the same. The intention will be to have all the three OS just run the exact same steps.

As a bonus, I've added a few recommended extensions for VS Code so anyone who might want to contribute will be off on the right foot (hopefully)


.github/workflows/cmake.yml
```diff
@@ -95,6 +93,29 @@ jobs:
...
+          version: "3.24.4"
+          url: "https://cmake.org/files/v3.24/cmake-3.24.4.tar.gz"
+      - uses: ./.github/actions/install/gtest
+
+      - name: setup
+        run: |
+          mkdir build
+          cd build
+          cmake .. -DJWT_BUILD_EXAMPLES=ON -DJWT_BUILD_TESTS=ON
+          sudo make install
```
Thalhammer, line 111: Is there a reason for the explicit make vs a `cmake --build . --target install` ?

.github/workflows/lint.yml
```diff
@@ -79,16 +75,14 @@ jobs:
...
-          sudo apt-get install clang-format-14
       - uses: ./.github/actions/render/defaults
         id: render
         with:
           traits_name: ${{ matrix.traits.name }}
           library_name: ${{ matrix.traits.library }}
           library_url: ${{ matrix.traits.url }}
           disable_default_traits: ${{ matrix.traits.disable_pico }}
-      - run: clang-format-14 -i ${{ steps.render.outputs.file_path }}
+      - run: clang-format -i ${{ steps.render.outputs.file_path }}
```
Thalhammer, line 85: I'd prefer sticking to a specific version, because when the version changes it often changes how existing code is formatted. Though I agree that we should probably update it to a more recent version given clang18 is already released.
It also serves as documentation to allow contributors to use the same version before commiting.

Thalhammer (approved):
Just some minor comments, otherwise good to merge from my side.

.github/workflows/cmake.yml
prince-chrismc: most of the are `make install` but there's no reason... it should really be `cmake --install .` for the ones not using presets

.github/workflows/lint.yml
prince-chrismc: I suspect it was left over from me debugging so let's see if it passes (which it should)

PR #352, opened 2024-06-05
Fix build and tests for wolfSSL

julek-wolfssl:
Tested with wolfSSL master
```
cmake -B build -DJWT_SSL_LIBRARY:STRING=wolfSSL -DJWT_BUILD_TESTS=ON  .
make -j -C build
./build/tests/jwt-cpp-test
```

### DIFF
--- a/.github/workflows/ssl.yml
+++ b/.github/workflows/ssl.yml
@@ -85,7 +85,6 @@ jobs:
   wolfssl:
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
         wolfssl:
           - { ref: "v5.1.1-stable", name: "5.1.1"}
--- a/.github/workflows/ssl.yml
+++ b/.github/workflows/ssl.yml
@@ -88,6 +88,7 @@ jobs:
           - { ref: "v5.1.1-stable", name: "5.1.1"}
           - { ref: "v5.2.0-stable", name: "5.2.0" }
           - { ref: "v5.3.0-stable", name: "5.3.0"}
+          - { ref: "v5.7.0-stable", name: "5.7.0"}
     name: wolfSSL ${{ matrix.wolfssl.name }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ssl.yml
+++ b/.github/workflows/ssl.yml
@@ -27,10 +27,10 @@ jobs:
           version: ${{ matrix.openssl.tag }}
 
       - name: configure
-        run: cmake . -DJWT_BUILD_TESTS=ON -DOPENSSL_ROOT_DIR=/tmp
-      - run: make
+        run: cmake --preset unit-tests -DOPENSSL_ROOT_DIR=/tmp
+      - run: cmake --build --preset unit-tests
       - name: test
-        run: ./tests/jwt-cpp-test
+        run: ctest --preset unit-tests
 
       - if: github.event_name == 'push' && always()
         uses: ./.github/actions/badge
@@ -50,8 +50,10 @@ jobs:
           version: "openssl-3.0.5"
 
       - name: configure
-        run: cmake . -DJWT_BUILD_TESTS=ON -DOPENSSL_ROOT_DIR=/tmp -DCMAKE_CXX_FLAGS="-DOPENSSL_NO_DEPRECATED=1" -DCMAKE_C_FLAGS="-DOPENSSL_NO_DEPRECATED=1"
-      - run: make
+        run: cmake --preset unit-tests -DOPENSSL_ROOT_DIR=/tmp -DCMAKE_CXX_FLAGS="-DOPENSSL_NO_DEPRECATED=1" -DCMAKE_C_FLAGS="-DOPENSSL_NO_DEPRECATED=1"
+      - run: cmake --build --preset unit-tests
+      - name: test
+        run: ctest --preset unit-tests
 
   libressl:
     runs-on: ubuntu-latest
@@ -69,10 +71,10 @@ jobs:
           version: ${{ matrix.libressl }}
 
       - name: configure
-        run: cmake . -DJWT_BUILD_TESTS=ON -DJWT_SSL_LIBRARY:STRING=LibreSSL
-      - run: make
+        run: cmake --preset unit-tests -DJWT_SSL_LIBRARY:STRING=LibreSSL
+      - run: cmake --build --preset unit-tests
       - name: test
-        run: ./tests/jwt-cpp-test
+        run: ctest --preset unit-tests
 
       - if: github.event_name == 'push' && always()
         uses: ./.github/actions/badge
@@ -100,10 +102,10 @@ jobs:
           version: ${{ matrix.wolfssl.ref }}
 
       - name: configure
-        run: cmake . -DJWT_BUILD_TESTS=ON -DJWT_SSL_LIBRARY:STRING=wolfSSL
-      - run: make
+        run: cmake --preset unit-tests -DJWT_SSL_LIBRARY:STRING=wolfSSL
+      - run: cmake --build --preset unit-tests
       - name: test
-        run: ./tests/jwt-cpp-test
+        run: ctest --preset unit-tests
 
       - if: github.event_name == 'push' && always()
         uses: ./.github/actions/badge
--- a/.github/workflows/ssl.yml
+++ b/.github/workflows/ssl.yml
@@ -83,6 +83,7 @@ jobs:
   wolfssl:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         wolfssl:
           - { ref: "v5.1.1-stable", name: "5.1.1"}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,7 @@ if(JWT_BUILD_EXAMPLES)
 endif()
 
 if(JWT_BUILD_TESTS)
+  enable_testing()
   add_subdirectory(tests)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
   # This is required to access OpenSSL compatibility API
   target_include_directories(jwt-cpp INTERFACE ${wolfssl_INCLUDE_DIRS})
   # EXTERNAL_OPTS_OPENVPN is necessary so that wolfssl/options.h is included automatically
-  target_compile_definitions(jwt-cpp INTERFACE OPENSSL_EXTRA OPENSSL_ALL EXTERNAL_OPTS_OPENVPN)
+  target_compile_definitions(jwt-cpp INTERFACE EXTERNAL_OPTS_OPENVPN)
 endif()
 
 if(NOT JWT_DISABLE_PICOJSON AND JWT_EXTERNAL_PICOJSON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,7 +125,8 @@ if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
   target_link_libraries(jwt-cpp INTERFACE PkgConfig::wolfssl)
   # This is required to access OpenSSL compatibility API
   target_include_directories(jwt-cpp INTERFACE ${wolfssl_INCLUDE_DIRS})
-  target_compile_definitions(jwt-cpp INTERFACE OPENSSL_EXTRA OPENSSL_ALL)
+  # EXTERNAL_OPTS_OPENVPN is necessary so that wolfssl/options.h is included automatically
+  target_compile_definitions(jwt-cpp INTERFACE OPENSSL_EXTRA OPENSSL_ALL EXTERNAL_OPTS_OPENVPN)
 endif()
 
 if(NOT JWT_DISABLE_PICOJSON AND JWT_EXTERNAL_PICOJSON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,18 +115,20 @@ target_include_directories(jwt-cpp INTERFACE $<BUILD_INTERFACE:${JWT_INCLUDE_PAT
 
 if(${JWT_SSL_LIBRARY} MATCHES "OpenSSL")
   target_link_libraries(jwt-cpp INTERFACE OpenSSL::SSL OpenSSL::Crypto)
-  target_compile_definitions(jwt-cpp INTERFACE JWT_OPENSSL)
 endif()
 
 if(${JWT_SSL_LIBRARY} MATCHES "LibreSSL")
   target_link_libraries(jwt-cpp INTERFACE LibreSSL::TLS)
-  target_compile_definitions(jwt-cpp INTERFACE JWT_LIBRESSL)
 endif()
 
 if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
   target_link_libraries(jwt-cpp INTERFACE PkgConfig::wolfssl)
   # This is required to access OpenSSL compatibility API
   target_include_directories(jwt-cpp INTERFACE ${wolfssl_INCLUDE_DIRS})
+  # This flag is required to have the manditory header included automatically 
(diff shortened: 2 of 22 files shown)


CMakeLists.txt
```diff
@@ -125,7 +125,8 @@ if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
   target_link_libraries(jwt-cpp INTERFACE PkgConfig::wolfssl)
   # This is required to access OpenSSL compatibility API
   target_include_directories(jwt-cpp INTERFACE ${wolfssl_INCLUDE_DIRS})
-  target_compile_definitions(jwt-cpp INTERFACE OPENSSL_EXTRA OPENSSL_ALL)
+  # EXTERNAL_OPTS_OPENVPN is necessary so that wolfssl/options.h is included automatically
```
prince-chrismc, line 128: Why was this header not required before? Would it be possible to just at the header?

tests/OpenSSLErrorTest.cpp
```diff
@@ -63,9 +63,15 @@ static uint64_t fail_PEM_write_bio_RSA_PUBKEY = 0;
 static uint64_t fail_RSA_set0_key = 0;
 #endif
 
+#ifdef LIBWOLFSSL_VERSION_STRING
+#define SYMBOL_NAME(s) ("wolfSSL_" s)
```
prince-chrismc, line 67: Do the plain openssl names not work? That seems like a breaking change 🤔

prince-chrismc (commented):
Thanks a ton for this! https://github.com/prince-chrismc/jwt-cpp/pull/35#issuecomment-2106331073 I've been staring at thos for a while so I really appreciate this.

I have a couple questions since I don't use wolfssl anymore I don't understand all the changes

CMakeLists.txt
julek-wolfssl: It was always necessary to include `wolfssl/options.h`. Notice that in your current builds you have
```
In file included from /usr/local/include/wolfssl/openssl/bn.h:33,
                 from /usr/local/include/wolfssl/openssl/ec.h:27,
                 from /home/runner/work/jwt-cpp/jwt-cpp/include/jwt-cpp/jwt.h:15,
                 from /home/runner/work/jwt-cpp/jwt-cpp/tests/JwksTest.cpp:1:
/usr/local/include/wolfssl/wolfcrypt/settings.h:2369:14: warning: #warning "For timing resistance / side-channel attack prevention consider using harden options" [-Wcpp]
 2369 |             #warning "For timing resistance / side-channel attack prevention consider using harden options"
      |              ^~~~~~~
```
which is a dead giveaway that `wolfssl/options.h` is missing. I think by defining `OPENSSL_EXTRA OPENSSL_ALL` you were able to somewhat build with wolfSSL but I don't know how the tests are "passing". I didn't investigate this.

tests/OpenSSLErrorTest.cpp
julek-wolfssl: We never exposed the same symbols as OpenSSL in our binary. I'm not sure how the tests pass without this PR. As soon as I fixed the `wolfssl/options.h` issue above this file blew up with errors and the functions would return NULL. Then the higher levels would try to call NULL which obviously segfaulted. `gdb` was very helpful to debug this:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00005555556278a8 in wolfSSL_BIO_new (type=0x7ffff7e8a440 <meth>)
    at /home/julek/Projects/jwt-cpp/tests/OpenSSLErrorTest.cpp:80
#2  0x00007ffff7db7e0e in wolfSSL_BIO_new_mem_buf (buf=0x555555743af0, len=2118) at ./src/bio.c:2639
#3  0x00005555555ab762 in jwt::helper::make_mem_buf_bio (
    data="-----BEGIN CERTIFICATE-----\nMIIF8DCCBVmgAwIBAgIKYFOB9QABAACIvTANBgkqhkiG9w0BAQUFADBGMQswCQYD\nVQQGEwJVUzETMBEGA1UEChMKR29vZ2xlIEluYzEiMCAGA1UEAxMZR29vZ2xlIElu\ndGVybmV0IEF1dGhvcml0eTAeFw0xMzA1MjIxNTQ5MD"...)
    at /home/julek/Projects/jwt-cpp/include/jwt-cpp/jwt.h:507
#4  0x00005555555ad2ea in jwt::helper::extract_pubkey_from_cert<jwt::error::rsa_error> (
    certstr="-----BEGIN CERTIFICATE-----\nMIIF8DCCBVmgAwIBAgIKYFOB9QABAACIvTANBgkqhkiG9w0BAQUFADBGMQswCQYD\nVQQGEwJVUzETMBEGA1UEChMKR29vZ2xlIEluYzEiMCAGA1UEAxMZR29vZ2xlIElu\ndGVybmV0IEF1dGhvcml0eTAeFw0xMzA1MjIxNTQ5MD"..., pw="", 
    ec=std::error_code = {std::_V2::error_category: 0}) at /home/julek/Projects/jwt-cpp/include/jwt-cpp/jwt.h:541
#5  0x00005555555ac9ee in jwt::helper::extract_pubkey_from_cert<jwt::error::rsa_error> (
    certstr="-----BEGIN CERTIFICATE-----\nMIIF8DCCBVmgAwIBAgIKYFOB9QABAACIvTANBgkqhkiG9w0BAQUFADBGMQswCQYD\nVQQGEwJVUzETMBEGA1UEChMKR29vZ2xlIEluYzEiMCAGA1UEAxMZR29vZ2xlIElu\ndGVybmV0IEF1dGhvcml0eTAeFw0xMzA1MjIxNTQ5MD"..., pw="")
    at /home/julek/Projects/jwt-cpp/include/jwt-cpp/jwt.h:578
#6  0x00005555555a7b9a in HelperTest_Cert2Pubkey_Test::TestBody (this=0x5555557516a0)
    at /home/julek/Projects/jwt-cpp/tests/HelperTest.cpp:11
#7  0x000055555568e4ff in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#8  0x0000555555682de6 in testing::Test::Run() ()
#9  0x0000555555682f65 in testing::TestInfo::Run() ()
#10 0x0000555555683519 in testing::TestSuite::Run() ()
#11 0x0000555555683c1f in testing::internal::UnitTestImpl::RunAllTests() ()
#12 0x000055555568eac7 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#13 0x000055555568302c in testing::UnitTest::Run() ()
#14 0x00005555555b0811 in RUN_ALL_TESTS () at /usr/include/gtest/gtest.h:2490
#15 0x00005555555b078a in main (argc=1, argv=0x7fffffffdd88) at /home/julek/Projects/jwt-cpp/tests/TestMain.cpp:5
```

prince-chrismc: Yikes. Well I am glad you knew to fox that since it was not obvious. I had hit bugs in the passed with the certificates and they were populated so I hadn't suspected anything 😅 

It might be worth adding an install of the cmake config files or maybe add these flags to the Pkg-config files?

julek-wolfssl: Which flags exactly do you mean? You can't use wolfSSL with different flags than it was compiled with since that will run into all kinds of trouble. A good solution would be to embed the defines inside of the `jwt.h` header but I'm not a cmake expert to make that happen.

Thalhammer:
Great to see someone from inside wolfSSL taking the time to improve support in opensource libraries.

I noticed the symbol names for wolfSSL are different from openssl (hence the new `SYMBOL_NAME` macro), which I assume are then converted to the OpenSSL ones using a define. The OpenSSL Error test works by providing a definition for the weakly defined symbols in openssl (which is also why it doesn't work with static linking). Given wolfssl has different symbol names, shouldn't the naming of the methods change as well, not just the methods they resolve to ? Keep in mind I have never worked with wolfSSL before, so I might be talking nonsense.

julek-wolfssl:
> Given wolfssl has different symbol names, shouldn't the naming of the methods change as well, not just the methods they resolve to ?

Our compatibility layer header are a set of macros that change the OpenSSL names to wolfSSL compatibility equivalents. Like `SSL_new` -> `wolfSSL_new` (`#define SSL_new wolfSSL_new` inside `wolfssl/openssl/ssl.h`.

tests/OpenSSLErrorTest.cpp
prince-chrismc: `EXTERNAL_OPTS_OPENVPN` for example.
No, the library should export the defines it was built with. Or else downstream will not know. End-users could build wolfssl with any options they like so it become problematic to hardcoded build settings here.

julek-wolfssl: That is the purpose of `wolfssl/options.h`. It contains the options used while building. `EXTERNAL_OPTS_OPENVPN` is only a convenience macro for compatibility users to not have to include `wolfssl/options.h`. Perhaps a better way would be to include `wolfssl/options.h` in `jwt.h`.

prince-chrismc: Ahhh. I that would make more sense. The OENVPN flag doesn't carry the same meaning.

julek-wolfssl: Its poor unfortunate legacy naming I agree. I'll try to refactor this patch.

julek-wolfssl: I added `options.h` including explicitly in the last commit.

julek-wolfssl:
```
/home/runner/work/jwt-cpp/jwt-cpp/tests/OpenSSLErrorTest.cpp:340:1: error: ‘RSA’ does not name a type
  340 | RSA* EVP_PKEY_get1_RSA(EVP_PKEY* pkey) {
      | ^~~
```
This is failing in the OpenSSL without deprecated functions action. I think the solution would be for `EVP_PKEY_get1_RSA` to not be used when building against OpenSSL with deprecated functions removed. But I also don't think that this is the PR to do this in.

prince-chrismc (commented):
This is so amazing helpful. Sorry for pushing the limits of the compatibility layer 🤣 but I see https://github.com/wolfSSL/wolfssl/pull/7618 so hopefully it was not too bad.

The only change I am unsure about is the new defines added to CMake, I think that could be avoided, the openssl and libressl are not used --- it makes it inconsistent with how it was being referenced.

I have no issue with disabling the tests that are not support or have conflicts. The rest of the changes make sense but I did have a few questions just to confirm.

Thanks again for helping to improve this.

tests/OpenSSLErrorTest.cpp
```diff
@@ -328,9 +337,9 @@ ECDSA_SIG* ECDSA_SIG_new(void) {
 		return origMethod();
 }
 
-struct rsa_st* EVP_PKEY_get1_RSA(EVP_PKEY* pkey) {
-	static struct rsa_st* (*origMethod)(EVP_PKEY * pkey) = nullptr;
-	if (origMethod == nullptr) origMethod = (decltype(origMethod))dlsym(RTLD_NEXT, "EVP_PKEY_get1_RSA");
+RSA* EVP_PKEY_get1_RSA(EVP_PKEY* pkey) {
```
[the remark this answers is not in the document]
prince-chrismc: This function is not even used 🤔  https://github.com/Thalhammer/jwt-cpp/blob/f9d4940da13c3888bff703a4385c204e2f3026b3/tests/OpenSSLErrorTest.cpp#L340

julek-wolfssl:
I appreciate the review. I will try to get back to this PR next week.

.github/workflows/ssl.yml
```diff
@@ -85,11 +85,13 @@ jobs:
   wolfssl:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
prince-chrismc, line 88:
```suggestion
```

CMakeLists.txt
```diff
@@ -115,17 +115,18 @@ target_include_directories(jwt-cpp INTERFACE $<BUILD_INTERFACE:${JWT_INCLUDE_PAT
 
 if(${JWT_SSL_LIBRARY} MATCHES "OpenSSL")
   target_link_libraries(jwt-cpp INTERFACE OpenSSL::SSL OpenSSL::Crypto)
+  target_compile_definitions(jwt-cpp INTERFACE JWT_OPENSSL)
```
prince-chrismc, line 118:
```suggestion
```

CMakeLists.txt
```diff
@@ -115,17 +115,18 @@ target_include_directories(jwt-cpp INTERFACE $<BUILD_INTERFACE:${JWT_INCLUDE_PAT
...
 endif()
 
 if(${JWT_SSL_LIBRARY} MATCHES "LibreSSL")
   target_link_libraries(jwt-cpp INTERFACE LibreSSL::TLS)
+  target_compile_definitions(jwt-cpp INTERFACE JWT_LIBRESSL)
```
prince-chrismc, line 123:
```suggestion
```

docs/install.md
```diff
@@ -30,6 +30,10 @@ The minimum is `jwt.h` but you will need to add the defines:
 
 - [`JWT_DISABLE_BASE64`](https://github.com/Thalhammer/jwt-cpp/blob/c9a511f436eaa13857336ebeb44dbc5b7860fe01/include/jwt-cpp/jwt.h#L11)
 - [`JWT_DISABLE_PICOJSON`](https://github.com/Thalhammer/jwt-cpp/blob/c9a511f436eaa13857336ebeb44dbc5b7860fe01/include/jwt-cpp/jwt.h#L4)
+- The SSL library being used needs to be defined:
+  - `JWT_OPENSSL` for OpenSSL (Default)
+  - `JWT_LIBRESSL` for LibreSSL
+  - `JWT_WOLFSSL` for wolfSSL
```
prince-chrismc, line 36:
```suggestion
```

CMakeLists.txt
```diff
@@ -115,17 +115,18 @@ target_include_directories(jwt-cpp INTERFACE $<BUILD_INTERFACE:${JWT_INCLUDE_PAT
...
 endif()
 
 if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
   target_link_libraries(jwt-cpp INTERFACE PkgConfig::wolfssl)
   # This is required to access OpenSSL compatibility API
   target_include_directories(jwt-cpp INTERFACE ${wolfssl_INCLUDE_DIRS})
```
prince-chrismc, line 129:
```suggestion
  target_include_directories(jwt-cpp INTERFACE ${wolfssl_INCLUDE_DIRS})
  # This flag is required to have the manditory header included automatically 
  # https://github.com/Thalhammer/jwt-cpp/pull/352#discussion_r1627971786
  #  https://github.com/wolfSSL/wolfssl/blob/3b74a6402998a8b8839e25e31ba8ac74749aa9b0/wolfssl/wolfcrypt/settings.h#L58
  target_compile_definitions(jwt-cpp INTERFACE EXTERNAL_OPTS_OPENVPN)
```

prince-chrismc (commented):
playing with getting the right headers pulled in

CMakeLists.txt
```diff
@@ -125,7 +125,10 @@ if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
...
-  target_compile_definitions(jwt-cpp INTERFACE OPENSSL_EXTRA OPENSSL_ALL)
+  # This flag is required to have the manditory header included automatically
```
prince-chrismc, line 128:
```suggestion
  # This flag is required to have the mandatory header included automatically 
```

prince-chrismc:
🤞 this should be ready. Thanks so much for contributing to both side to make the integration better 🚀 

I wont be updating the docs with the new tested version because I have a PR to update the the SSL versions which would conflict.

julek-wolfssl:
Thanks for helping out here. I haven't had time to bring it across the finish line in the past month.

PR #357, opened 2024-07-25
Jsoncpp traits no longer uses deprecated Reader

kiddigital:
The jsoncpp trait made use of the [deprecated Reader](https://github.com/open-source-parsers/jsoncpp/releases/tag/1.4.0).

This PR fixes this. Safer and no more deprecation warnings when compiling.

### DIFF
--- a/include/jwt-cpp/traits/open-source-parsers-jsoncpp/traits.h
+++ b/include/jwt-cpp/traits/open-source-parsers-jsoncpp/traits.h
@@ -120,10 +120,8 @@ namespace jwt {
 				Json::CharReaderBuilder builder;
 				const std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
 
-				return reader->parse(
-					reinterpret_cast<const char*>(str.c_str()),
-					reinterpret_cast<const char*>(str.c_str() + str.size()), &val,
-					nullptr);
+				return reader->parse(reinterpret_cast<const char*>(str.c_str()),
+									 reinterpret_cast<const char*>(str.c_str() + str.size()), &val, nullptr);
 			}
 
 			static string_type serialize(const value_type& val) {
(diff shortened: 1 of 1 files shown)


prince-chrismc:
1.4 when we are using 1.9 I don't think it's a breaking change to requie this but it's worth noting.

kiddigital:
> 1.4 when we are using 1.9 I don't think it's a breaking change to requie this but it's worth noting.

It has been marked as Deprecated since 1.4 😁
So quite likely that the deprecated Reader gets dropped some time soon in an upcoming version of jsoncpp.
And also very likely that people that are using jsoncpp (in combination with jwt-cpp) will use a more recent version of jsoncpp. At least 1.4.

Who doesn't want to get rid of these deprecated messages. And when incorporating something new like jwt-cpp, getting these deprecated messages does not _bring joy_ 😉

prince-chrismc:
I see only 1 linter failed, https://github.com/Thalhammer/jwt-cpp/actions/runs/10100944975/job/27938470947?pr=357#step:5:82

If you don't there's instructions to fix it quickly, of not I will do my best to make time to push it myself.

prince-chrismc (approved):

kiddigital:
> I see only 1 linter failed, ...

I saw that as well, but it is the 'process linter results' step 😁

Not something I think I can/should fix. Or did I overlook something?

prince-chrismc:
It's telling you there are changes to be applied to the PR 😜  again it tells you the exact command to run to fix it sonics just a copy paste... the code diff is there too if you expand it.

kiddigital:
Ah.. ok.. I need to read better 😜 Will have a look asap and apply the fixes...

Thalhammer (approved):
Seems good to me 👍

kiddigital:
@prince-chrismc , you just beat me to it 😁 thx

prince-chrismc:
My pleasure :) Thanks for the contribution! 🚀

PR #361, opened 2024-09-21
Ensure array_type has front() method + fix array_type on jsoncpp

omar-mohamed-khallaf:
Replace `front()` with `[0]` in `get_x5c_key_value` as front is not supported by jsoncpp on array types.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -3925,7 +3925,7 @@ namespace jwt {
 			auto x5c_array = get_jwk_claim("x5c").as_array();
 			if (x5c_array.size() == 0) throw error::claim_not_present_exception();
 
-			return json_traits::as_string(x5c_array.front());
+			return json_traits::as_string(x5c_array[0]);
 		};
 
 		/**
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -2459,9 +2459,7 @@ namespace jwt {
 		struct is_valid_json_array {
 			template<typename T>
 			using value_type_t = typename T::value_type;
-			using front_return_type = decltype(std::declval<array_type>().front());
-			using front_base_type =
-				typename std::remove_cv<typename std::remove_reference<front_return_type>::type>::type;
+			using front_base_type = typename std::decay<decltype(std::declval<array_type>().front())>::type;
 
 			static constexpr auto value = std::is_constructible<value_type, array_type>::value &&
 										  is_iterable<array_type>::value &&
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -3925,7 +3925,7 @@ namespace jwt {
 			auto x5c_array = get_jwk_claim("x5c").as_array();
 			if (x5c_array.size() == 0) throw error::claim_not_present_exception();
 
-			return json_traits::as_string(x5c_array[0]);
+			return json_traits::as_string(x5c_array.front());
 		};
 
 		/**
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -808,7 +808,7 @@ namespace jwt {
 		 * \brief Load a public key from a string.
 		 *
 		 * The string should contain a pem encoded certificate or public key
-		 * 
+		 *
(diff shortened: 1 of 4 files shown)


prince-chrismc:
Would it be possible to implement the missing function in the trait definition?

Extending,

https://github.com/Thalhammer/jwt-cpp/blob/83703d718a180b178bfecac0f12355358d04181b/include/jwt-cpp/traits/open-source-parsers-jsoncpp/traits.h#L16

Also it seems we are missing this function as a requirement for the json array

https://github.com/Thalhammer/jwt-cpp/blob/83703d718a180b178bfecac0f12355358d04181b/include/jwt-cpp/jwt.h#L2459

omarmohamedkh:
Is `front()` the missing function for the second issue, or something else?

prince-chrismc:
Yes, `front()` Should be checked.

Might also be fore checking if the unit tests are missing  a case 🤔

omar-mohamed-kh:
> Yes, `front()` Should be checked.

I added this line `!std::is_same<decltype(std::declval<array_type>().front()), void>::value` to `is_valid_json_array`, is it ok?

`std::is_member_function_pointer` had problems resolving the overloaded `front` function, and I had no idea how to solve it.

include/jwt-cpp/jwt.h
```diff
@@ -2463,7 +2463,8 @@ namespace jwt {
 			static constexpr auto value = std::is_constructible<value_type, array_type>::value &&
 										  is_iterable<array_type>::value &&
 										  is_detected<value_type_t, array_type>::value &&
-										  std::is_same<typename array_type::value_type, value_type>::value;
+										  std::is_same<typename array_type::value_type, value_type>::value &&
+										  !std::is_same<decltype(std::declval<array_type>().front()), void>::value;
```
prince-chrismc, line 2467: Hmm this is a good method but I am curious is the return should be `value_type_t` rather them not void?

I generally see these written separately so the name is a bit more meaningful

https://github.com/Thalhammer/jwt-cpp/blob/feaca3cc8541e49b3508fdd07abe83851f858c91/include/jwt-cpp/jwt.h#L2430

include/jwt-cpp/traits/danielaparker-jsoncons/traits.h
```diff
@@ -61,7 +61,12 @@ namespace jwt {
 					return 0;
 				}
 			};
-			using array_type = json::array;
+			class array_type : public json::array {
+			public:
+				value_type& front() { return this->operator[](0U); }
```
prince-chrismc, line 66: Do we need the none const version?


We only ever call it with const from `as_string`
https://github.com/Thalhammer/jwt-cpp/blob/feaca3cc8541e49b3508fdd07abe83851f858c91/include/jwt-cpp/traits/open-source-parsers-jsoncpp/traits.h#L108

omar-mohamed-khallaf: `jsoncpp` supported both, so I thought no harm to add both and copied them to `jsoncons` as well.
I'll remove them.

include/jwt-cpp/jwt.h
omar-mohamed-khallaf: But `value_type_t` is templated, wouldn't this make `value` templated as well?
Really not sure how `is_detected` using it or what does it do.

Anyway, I changed it to `std::is_member_function_pointer` since I removed the non-const version.

prince-chrismc (approved):
Amazing work, thanks for helping to fix this!

prince-chrismc:
Looks like a type error in the Danieljson changes from the CI

omar-mohamed-khallaf:
Something is wrong in the current implementation.
```c++
static object_type as_object(const json& val) {
	if (val.type() != jsoncons::json_type::object_value) throw std::bad_cast();
	return object_type(val.object_value());
}

static array_type as_array(const json& val) {
	if (val.type() != jsoncons::json_type::array_value) throw std::bad_cast();
	return val.array_value();
}
```

These two functions use `val.object_value()` and `val.array_value()`, but I can't find any reference to these functions in the jsoncons library.

omar-mohamed-khallaf:
The problem existed before the current modifications.
Should it be fixed in a new PR?

PR #365, opened 2024-10-31
Update CMakeLists.txt

dantti:
Specify C++ language to avoid searching for C compiler


Thalhammer:
Looks reasonable to me, we do indeed not use any C so there's no harm in disabling C support.

PR #369, opened 2025-02-25
Support passing ssl library key handles to algorithms

sandro97git:
I have a use case where I already have the private key represented as an openssl EVP_PKEY instace. Currently an alogrithm constructor for passing such an object directly is missing.

### DIFF
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -1570,11 +1570,9 @@ namespace jwt {
 					throw error::ecdsa_exception(error::ecdsa_error::invalid_key_size);
 			}
 
-			ecdsa(helper::evp_pkey_handle private_key, const EVP_MD* (*md)(), std::string name, size_t siglen)
-				: pkey(std::move(private_key)), md(md), alg_name(std::move(name)), signature_length(siglen) {
-				if (pkey) {
-					check_private_key(pkey.get());
-				} else {
+			ecdsa(helper::evp_pkey_handle key_pair, const EVP_MD* (*md)(), std::string name, size_t siglen)
+				: pkey(std::move(key_pair)), md(md), alg_name(std::move(name)), signature_length(siglen) {
+				if (!pkey) {
 					throw error::ecdsa_exception(error::ecdsa_error::no_key_provided);
 				}
 				size_t keysize = EVP_PKEY_bits(pkey.get());
--- a/include/jwt-cpp/jwt.h
+++ b/include/jwt-cpp/jwt.h
@@ -1461,6 +1461,17 @@ namespace jwt {
 					throw error::rsa_exception(error::rsa_error::no_key_provided);
 			}
 			/**
+			 * Construct new rsa algorithm
+			 *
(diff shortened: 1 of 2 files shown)


include/jwt-cpp/jwt.h
```diff
@@ -1570,6 +1570,18 @@ namespace jwt {
 					throw error::ecdsa_exception(error::ecdsa_error::invalid_key_size);
 			}
 
+			ecdsa(helper::evp_pkey_handle private_key, const EVP_MD* (*md)(), std::string name, size_t siglen)
```
sandro97git, line 1573: I currently only added the minimal code required to make my use case work. Please let me know if passing an EVP_PKEY directly is something that this library is willing to support in the future.
If yes, we can think about how to fully implement this feature. I would probably also need to add an rsa implementation...

Thalhammer:
Adding support for this is fine from my side. We might wanna consider adding a few things though:
- Like you already said, a RSA variant is missing
- We should probably support passing a public key as well (pkey can store both public and private keys) to allow using it when only verification is required

You can ignore the CI, its failing because I forgot to update the upload-artifact action.

sandro97git:
I added commit [e37e259](https://github.com/Thalhammer/jwt-cpp/pull/369/commits/e37e25906abca189a72eb3276575af62ad338649)  for supporting public keys as well

Thalhammer:
Looks fine so far to me.

sandro97git:
@Thalhammer Do you know why the coverage check is failing?

prince-chrismc:
> @Thalhammer Do you know why the coverage check is failing?

Seems to be having an issue with the gtest code gen from the macro 🤔

prince-chrismc (approved):
Think out load... I am curious if this is more error prone? you can pass any key with any algorithm and name. 

Bonus points, new PR with a negative test for the MD* being wrong to the size that would be interesting to see how it fails 🏗️ 

otherwise this matchs our exsting ctor with a little duplication 👍

sandro97git:
I agree this is indeed a bit error prone. However this is already possible using the base `ecdsa` or `rsa` constructor.
I'm not an expert on this jwt algorithms but I wonder if it wouldn't be possible to select the alorithm solely based on the key length. Something similar to this code:
```c++
const auto bits = EVP_PKEY_bits(private_key.GetRaw());
const EVP_MD* (*md)();
std::size_t siglen;
std::string name;
switch (bits) {
  case 384:
    md = EVP_sha384;
    siglen = 96;
    name = "ES384";
    break;
  case 521: {
    md = EVP_sha512;
    siglen = 132;
    name = "ES512";
    break;
  }
  default:
    throw std::invalid_argument{"unsupported key size: " + std::to_string(bits)};
}
```

Would this be possible or are keys of the same lenght used for multiple algorithms?

Thalhammer:
> @Thalhammer Do you know why the coverage check is failing?

Not yet, but I assume it might be because of old versions. I'll look into it at some point, its not a mandatory check so theres not much harm in ignoring it for now.

PR #387, opened 2025-06-19
Fix error in CMake config-file package

eaaltonen:
### Commit

* Remove greater than sign from link interface

Fixes error
```
  The link interface of target "jwt-cpp::jwt-cpp" contains:

    picojson::picojson>

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ if(${JWT_SSL_LIBRARY} MATCHES "wolfSSL")
 endif()
 
 if(NOT JWT_DISABLE_PICOJSON AND JWT_EXTERNAL_PICOJSON)
-  target_link_libraries(jwt-cpp INTERFACE picojson::picojson>)
+  target_link_libraries(jwt-cpp INTERFACE picojson::picojson)
 endif()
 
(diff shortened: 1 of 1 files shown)


eaaltonen:
> What version of CMake

@prince-chrismc : i observed this with CMake versions 4.0.3 ( _from conda-forge in_ `conda-build`) and CMake version 3.31.6 bundled with CLion.

> how are you generating the Config for picojson?

This is based on my original [PR to picojson](https://github.com/kazuho/picojson/pull/159) which received no reply; Since
* `picojson::picojson` would pretty much be the default target name for picojson in CMake
* this repos CMake already had a reference to the target (_minus the_ >)

we recently merged a change to https://github.com/conda-forge/picojson-feedstock that creates a CMake config-file package. It is now included in Conda [picojson packages](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fpicojson-1.3.0-h84d6215_1.conda) build versions 1.

PR #391, opened 2025-08-10
glz::json_t trait support

StormLord07:
* added a trait support for glz::json_t, map like glaze object jwt::traits::stephenberry_glaze
* and a default.h file

### DIFF
--- a/include/jwt-cpp/traits/stephenberry-glaze/defaults.h
+++ b/include/jwt-cpp/traits/stephenberry-glaze/defaults.h
@@ -1,0 +1,91 @@
+#ifndef JWT_CPP_STEPHENBERRY_GLAZE_DEFAULTS_H
+#define JWT_CPP_STEPHENBERRY_GLAZE_DEFAULTS_H
+
+#ifndef JWT_DISABLE_PICOJSON
+#define JWT_DISABLE_PICOJSON
+#endif
+
+#include "traits.h"
+
+namespace jwt {
+	/**
+	 * \brief a class to store a generic [glz::json_t](https://github.com/stephenberry/glaze/) value as claim
+	 *
+	 * This type is the specialization of the \ref basic_claim class which
+	 * uses the standard template types.
+	 */
+	using claim = basic_claim<traits::stephenberry_glaze>;
+
+	/**
+	 * Create a verifier using the default clock
+	 * \return verifier instance
+	 */
+	inline verifier<default_clock, traits::stephenberry_glaze> verify() {
+		return verify<default_clock, traits::stephenberry_glaze>(default_clock{});
+	}
+
+	/**
+	 * Create a builder using the default clock
+	 * \return builder instance to create a new token
+	 */
+	inline builder<default_clock, traits::stephenberry_glaze> create() {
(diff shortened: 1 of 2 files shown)


Thalhammer:
Hi and thank you for showing interest in extending jwt-cpp's support for json libraries.
Would it be possible to add some CI testing to ensure we don't break any of the code in the future ?
In addition the CMake scripts need modification to ensure the new files are correctly installed along the other headers.

StormLord07:
@Thalhammer I’ve added tests. Glaze is usually installed via `FetchContent`, but it can be found using `find_package`. I used `find_package` like in other tests, though it can't be installed via apt and has to be built from source and installed manually

A few points to mention:

1. `glz::json_t` doesn’t store integer values, only doubles. I didn’t add special support for integers. The only workaround I can think of is checking if `double == floor(double)`, which feels wrong because we might want `16.0` to not be the same as `16`.
2. There’s no support for string streams and `operator>>` with `glz::json_t`, so I had to change the test code. We could either add artificial support for it or i could try adding PR to Glaze, but that would take some time.

StormLord07:
It should work, unfortunately i cant fix it without changing already exisiting ci too much, its just JWT CI coverage fails on 

```
geninfo: ERROR: mismatched end line for _ZN25BaseTest_Base64Index_Test8TestBodyEv at /home/runner/work/jwt-cpp/jwt-cpp/tests/BaseTest.cpp:4: 4 -> 20
	(use "geninfo --ignore-errors mismatch ..." to bypass this error)
```

the fix should be changinx tests CMake file with provding additional args to 
` setup_target_for_coverage_lcov(NAME coverage EXECUTABLE ${CMAKE_CURRENT_BINARY_DIR}/jwt-cpp-test)`

but 
* im not sure how do it properly
* its outside of scope of this pr,

All other problems were mentioned before

> 1. `glz::json_t` doesn’t store integer values, only doubles. I didn’t add special support for integers. The only workaround I can think of is checking if `double == floor(double)`, which feels wrong because we might want `16.0` to not be the same as `16`.
> 2. There’s no support for string streams and `operator>>` with `glz::json_t`, so I had to change the test code. We could either add artificial support for it or i could try adding PR to Glaze, but that would take some time.

tests/traits/StephenberryGlazeTest.cpp:19
tests/traits/StephenberryGlazeTest.cpp:58

Thalhammer:
> Glaze is usually installed via `FetchContent`

Using FetchContent is also fine, we already do that for nlohmann json. If you do that it would be good to add a option to disable it so that people can turn it off if they don't want to use it.

> so I had to change the test code.

That's perfectly fine for me. We don't guarantee/require any support from the json library except whats needed inside jwt-cpp or the trait itself.

> its just JWT CI coverage fails on

Yeah that's known. Coverage broke a while ago due to version changes in githubs runners, I just haven't come around to update it yet. It's outside the scope of the PR, so don't worry about it ;)
If you are bored you are free to fix it in a different PR though.

StormLord07:
Ah, that's what it means, I thought it just is unable to push into the repo, fixed.

DockedFerret800:
@Thalhammer Any chance this could be merged soon?

PR #393, opened 2025-08-18
Fixed CI/CD coverage

StormLord07:
I managed to fix the CI/CD pipeline for coverage by switching to gcovr, so if this is not something you want to do, tell me i'll try to find a solution to lcov, but i've tried updating versions and many other things and it still fails because of line endings, [like in this issue](https://github.com/linux-test-project/lcov/issues/192) but then it requires more and more ignores, at some point it doesn't look like coverage, here is the resulting link [coveralls](https://coveralls.io/jobs/170178997), 

the Lint is still failing, but it looks like its by design. so the changes would be applied only manually?

### DIFF
--- a/.github/workflows/jwt.yml
+++ b/.github/workflows/jwt.yml
@@ -23,7 +23,7 @@ jobs:
         run: cmake --preset coverage
         #      - name: run-lcov
         #        run: cmake --build --preset coverage
-      - name: run-gcovr
+      - name: run
         run: cmake --build --preset coverage
 
       - uses: coverallsapp/github-action@v2


prince-chrismc:
This use to work with no code changes. I suspect the problem is the gcc or llvm and lcov in the github runner are incompatible. In my testing they are different versions.

The output has been super consistent but its been a while since I've looked at this https://github.com/prince-chrismc/jwt-cpp/pull/48/commits/8c8369361a2e80d08a156bed2f3aface9aa225b2


Why did the cmake module being used not work, is has support for both tools?

StormLord07:
> This use to work with no code changes. I suspect the problem is the gcc or llvm and lcov in the github runner are incompatible. In my testing they are different versions.

Hmm, If i knew the commit at which the CI started failing, I'll go look into it, maybe the answer is in there, a lot of things started failing when we at our project updated gcc maybe here is something alike.

> Why did the cmake module being used not work, is has support for both tools?

I'll try it out, couldn't find the info on the gcovr, might have to research that bit, was pretty late when i was finishing things up.

prince-chrismc:
Seems the github action runner were updated and this has been unblocked.

PR #404, opened 2026-01-29
Remove dependency on pkg-config for wolfSSL

lapinozz:
pkg-config doesn't seem to be actually used for anything
it's simpler to use find_package() just like for the other libraries

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -37,6 +37,7 @@ body:
       label: Version
       description: What version of our software are you running?
       options:
+        - 0.7.2
         - 0.7.1
         - 0.7.0
         - 0.6.0
--- a/.github/actions/install/cmake/action.yml
+++ b/.github/actions/install/cmake/action.yml
@@ -7,15 +7,34 @@ inputs:
   url:
     description: "The corresponding URL to download the source code from"
     required: true
+  cache-key-extra:
+    description: "The key to use for caching the built CMake. Defaults to a key based on the version and runner architecture."
+    required: false
+    default: "host"
 runs:
   using: composite
   steps:
+  # Taken from https://github.com/actions/setup-go/blob/a5f9b05d2d216f63e13859e0d847461041025775/README.md?plain=1#L343-L365
+  - name: Normalize runner architecture (Linux/macOS)
+    if: runner.os != 'Windows'
+    shell: bash
+    run: echo "ARCH=$(echo '${{ runner.arch }}' | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
+  - name: Normalize runner architecture (Windows)
+    if: runner.os == 'Windows'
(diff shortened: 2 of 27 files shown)


lapinozz:
> What release was official cmake targets export by wolfssl? 

I don't understand the question

> Do this require an update to the CI for what versions are tested against?

I'm not sure, I haven't checked the CI, I only tested to compile locally in static and shared build

I would just be a nice improvement since pck-config is an additional dependency and VS/MSVC doesn't come with it

prince-chrismc:
https://github.com/wolfSSL/wolfssl/blob/ca5b484e23beb4c033eaf8a9f5261cb417f39b4e/CMakeLists.txt#L3112-L3116

Seems they do have official CMake targets added ~3years ago... we are a bit behind

prince-chrismc:
@lapinozz are you using the wolfSSL CMake install? Are you using wolfSSL on windows? 

having looked into myself, it's not mature enough shape for us to use it for jwt-cpp unfortunately.  It was completely broken until https://github.com/wolfSSL/wolfssl/commit/0319eb098d704342227efbe4d438234d0d0e62e6. And looking into the code we'd need to test against... There's a few critical TODOs in their code for enable crypto https://github.com/wolfSSL/wolfssl/blob/9e5d03b23ece0d872fe20c3b6a1f1ee28dbd9912/CMakeLists.txt#L238 features.

Lastly I found https://github.com/wolfSSL/wolfssl/commit/540b51eb28ec001c409aa61e86bbe1e860e4a210 which looks promising because it allows the autoconf to export the required CMake targets actually build with the correct set of  features. But this is completely unreleased and very ( 3 weeks ) new.

At this time I dont think we can remove the pkg-config requirement for wolfSSL.

prince-chrismc:
Notes for future me to recall what I tried

```
 9150  sudo rm -rf /tmp/wolfssl-*
 9151  '/mnt/c/Users/Chris McArthur/Documents/jwt-cpp/.github/actions/install/wolfssl/install-wolfssl.sh' v5.7.0-stable
 9152  ls /usr/local/lib
 9153  ls /usr/local/lib/cmake
 9154  ls /usr/local/lib/pkgconfig
 9155  cd /tmp/wolfssl-5.7.0-stable
 9156  mkdir build
 9157  cd build
 9158  cmake ..
 9159  make -j $(nproc)
 9160  make install
 9161  cd ~
 9162* '/mnt/c/Users/Chris McArthur/Documents/jwt-cpp/.github/actions/install/wolfssl/install-wolfssl.sh' v5.7.6-stable
 9163  cd /tmp/wolfssl-5.7.6-stable
 9164  mkdir build
 9165  cmake ..
 9166  cd build
 9167  cmake ..
 9168  make install
 9169  cat /usr/local/lib/cmake/wolfssl/wolfssl-targets.cmake
 9170  cmake .. -DWOLFSSL_OPENSSLEXTRA=yes -DWOLFSSL_OPENSSLALL=yes
 9171  make install
```

PR #405, opened 2026-02-07
Cast data size to unsigned int

JurgenLB:
this is to address some additional warnings for the type conversion.

4 EVP_DigestUpdate warnings (lines 1627, 1667, 1987, 2036):
static_cast<unsigned int>(data.size())
to explicitly convert size_t (__int64 on 64-bit Windows) to unsigned int

1 time conversion warning (line 2673):
static_cast<int64_t>(std::llround(as_number()))
to explicitly handle the conversion from long long to int64_t

These conversions are safe because JWT data is always small (< 4GB) or have the same range

kind regards


JurgenLB:
If you want I can separate the coverage report and the type conversions.

Why the coverage reports was not working I do not know.
For this to pass I only added `LCOV_ARGS "--ignore-errors;mismatch;--ignore-errors;unused")`

Kind regards

prince-chrismc (approved):

